PR #45, opened 2023-02-13
FLPATH-7: improve code coverage for model api

pkliczewski:
This PR improves code coverage to 81% and missed branches to 75%.

### DIFF
--- a/notification-service/src/lombok.config
+++ b/notification-service/src/lombok.config
@@ -1,0 +1,2 @@
+config.stopBubbling = true
+lombok.addLombokGeneratedAnnotation = true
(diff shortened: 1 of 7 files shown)


parodos-model-api/src/test/java/com/redhat/parodos/infrastructure/option/InfrastructureOptionsTest.java
```diff
@@ -0,0 +1,95 @@
...
+        options.addUpgradeOption(option);
+
+        assertTrue(options.getUpgradeOptions().size() == 1);
+    }
+
+    @Test
+    public void checkInfra() {
+        InfrastructureOptions options = new InfrastructureOptions.Builder().setCurrentInfrastructure(option).build();
+
+        assertFalse(options.hasInfrastructure());
```
pkliczewski, line 89: Is this correct?

```java
	public boolean hasInfrastructure() {
		return currentVersion == null;
	}
```

Either the name is confusing or we have a bug

lshannon: The name is confusing. This should be indicating if we were able to determine if the project already had a Workflow applied to it. What I mean by this is the project might be on OCP3 already. So in the UI we would display the project is on OCP3, and then perhaps give them an option to upgrade to OCP4

PR #54, opened 2023-02-16
FLPATH-5: Update and include Dockerfile and Docker Compose

anludke:


eloycoto (approved):

docker-compose/run-parodos-locally.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
eloycoto, line 1: I would add a comment in the Readme, but looks good to me :-)

anludke: Awesome! Thank you

PR #63, opened 2023-02-21
CI: add better integration on GitHub Actions.

eloycoto:
This change adds a way to test the following things in GitHub actions:

- Build using maven build as it was.
- A new job to report jacoco reports.
- A super simple way to deploy things to kubernetes for example.
- A simple way to run the example script.
- A big refactor on example script with colors and assertions.
- Added cache on maven when build

The workflow is like this one:
https://github.com/redhat-developer/parodos/actions/runs/4232739123

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,46 @@
+
+COMPOSE ?= docker-compose
+DOCKER ?= docker
+
+ORG=quay.io/parados/
+WORKFLOW_SERVICE_IMAGE=workflow-service
+NOTIFICATION_SERVICE_IMAGE=notification-service
+
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed s,^main$$,latest,g)
+GIT_HASH := $(shell git rev-parse HEAD)
+
+build-images:
+	$(COMPOSE) -f ./docker-compose/docker-compose.yml build
+
+
+tag-images:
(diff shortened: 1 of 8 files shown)


hack/manifests/base/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+resources:
```
pkliczewski, line 1: Do you think it makes sense to have more manifests? We have local profile but we need integration and production as well.

Makefile
```diff
@@ -0,0 +1,46 @@
+
+COMPOSE ?= docker-compose
+DOCKER ?= docker
+
+ORG=quay.io/parados/
+WORKFLOW_SERVICE_IMAGE=workflow-service
+NOTIFICATION_SERVICE_IMAGE=notification-service
+
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed s,^EloyGH$$,test,g)
```
pkliczewski, line 9: Do we need `EloyGH` here?

hack/manifests/base/kustomization.yaml
eloycoto: At the moment I added the testing folder, and from there we can evolve in case that it's needed ;-)

pkliczewski: OK, make sense

lshannon (approved):

PR #64, opened 2023-02-22
Test: Exclude lombok code from coverage

eloycoto:
Fix FLPATH-31


pkliczewski (commented):
According to this [blog](https://dev.to/derlin/exclude-lombok-generated-code-from-test-coverage-jacocosonarqube-4nh1) we can `put it either at the root of your repo, or in the src/ folder`. What do you think to have it at the root level?

pkliczewski (approved):

anludke (approved):

PR #70, opened 2023-02-23
Test: add unit test for the UserService

eloycoto:


workflow-service/src/test/java/com/redhat/parodos/user/service/UserServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+		// when
+		UserResponseDTO res = this.service.getUserByUsername(user.getUsername());
+
+		// then
+		assertNull(res);
+		Mockito.verify(this.userRepository, Mockito.times(1)).findByUsername(Mockito.any());
+	}
+
+	private User getSampleUser(String name) {
+		User user = User.builder().username(name).email("test@test.com").build();
```
pkliczewski, line 115: We pass the name but hardcode the email. What do you think about passing both or hardcoding both?

eloycoto: The idea is to have more than one, so you have a different one, if something change, u do something in the test itself, like:

user = getSampleUser("test)
userFoo = getSampleUser("foo")
userFoo.setEmail("test@foo.com")

So, the idea is to have lean helpers, that are used across all places

pkliczewski: my point was about consistency. We could either have something like:

```java
private User getSampleUser() {
		User user = User.builder().username("test").email("test@test.com").build();
		user.setId(UUID.randomUUID());
		return user;
}
```

or 

```java
private User getSampleUser(String name, String email) {
		User user = User.builder().username(name).email(email).build();
		user.setId(UUID.randomUUID());
		return user;
}
```

For current code it should be as easy as using the first one but if we think about broad reuse the second one seems to be better.

eloycoto: The thing that I wanted to avoid is writing the data twice, where is no value for that. because Name is something that we need for each test, but no the email at all.

workflow-service/src/test/java/com/redhat/parodos/user/service/UserServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+		// given
+		User user = getSampleUser("test");
+		Mockito.when(this.userRepository.save(Mockito.any(User.class))).thenReturn(user);
+
+		// when
+		UserResponseDTO res = this.service.save(user);
+
+		// then
+		assertNotNull(res);
+		assertEquals(res.getId().toString(), user.getId().toString());
```
gciavarrini, line 42: `res.getId()` returns `String` so you can simplify this assertion as
`assertEquals(res.getId(), user.getId().toString());`

workflow-service/src/test/java/com/redhat/parodos/user/service/UserServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+		// given
+		User user = getSampleUser("test");
+		Mockito.when(this.userRepository.findById(user.getId())).thenReturn(Optional.of(user));
+
+		// when
+		UserResponseDTO res = this.service.getUserById(user.getId());
+
+		// then
+		assertNotNull(res);
+		assertEquals(res.getId().toString(), user.getId().toString());
```
gciavarrini, line 60: Same as above

workflow-service/src/test/java/com/redhat/parodos/user/service/UserServiceImplTest.java
pkliczewski: What do you think about using the first option. We always use the same name.

eloycoto: at some point, test need to be extended where the findbyUserName is different and have more than one value in the array.

lshannon (approved):

PR #71, opened 2023-02-23
Fix README files typos

gciavarrini:
This PR fixed some typos present in README files.
In addition, It reformats README files using line length equals to 120 chars. This is necessary to ease the read of the raw file and to ease the edits on the text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,17 +10,17 @@ progress).
 https://github.com/janus-idp
 
 The focus of Parodos is around Workflows (composed of WorkflowTasks) that bring together the existing tools and
-processes of the enterpise in an end-to-end experience that developers, quality assurance, produnction support and other
+processes of the enterpise in an end-to-end experience that developers, quality assurance, production support and other
 enterprise software development/delivery team members can consume (through our Backstage plugins or by your own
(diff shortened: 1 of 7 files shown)


pkliczewski (commented):
Overall, it looks good. I only have a question about breaking single line of text to multiple lines. Is it about esthetics?

gciavarrini:
> Overall, it looks good. I only have a question about breaking single line of text to multiple lines. Is it about esthetics?

Fixing a typo when a line is, for example, 600 character long is a bit difficult because you have to scroll horizontally. So, IMO, wrapping lines is a good way to help the developer to read and fix markdown files quickly.

In addition, wrapping lines ease the review process: when fixing typos on a short line, it's faster to see the differences from `git diff` (see [example](https://about.gitlab.com/blog/2016/10/11/wrapping-text/#comparing-similar-views))

As far as I know, there is no a de facto best practice about line wrap in Markdown, but for example the [GitLab style guide](https://docs.gitlab.com/ee/development/documentation/styleguide/#text) imposes the line wrapping.

@pkliczewski WDYT?

pkliczewski:
It is ok to me. Let's see what others will say

lshannon (approved):

PR #72, opened 2023-02-27
FLPATH-44: workflow level parameters

RichardW98:
[FLPATH-44](https://issues.redhat.com/browse/FLPATH-44)

Screenshot 2023-02-26 at 10 12 47 PM


workflow-examples/src/main/java/com/redhat/parodos/examples/simple/LoggingWorkFlowTask.java
```diff
@@ -43,16 +45,30 @@ public class LoggingWorkFlowTask extends BaseInfrastructureWorkFlowTask {
...
+		return new DefaultWorkReport(WorkStatus.FAILED, workContext);
 	}
 
 	@Override
 	public List<WorkFlowTaskParameter> getWorkFlowTaskParameters() {
-		return List.of(WorkFlowTaskParameter.builder().key("api-server").description("The api server")
-				.type(WorkFlowTaskParameterType.URL).optional(false).build());
+		return List.of(
+				WorkFlowTaskParameter.builder().key("api-server").description("The api server")
+						.type(WorkFlowTaskParameterType.URL).optional(false).build(),
```
pkliczewski, line 68: How do we differentiate between optional and missing parameter and required and missing? Do we rely on user to verify whether it was provided?

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/parameter/WorkFlowTaskParameterScope.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * The supported scopes for an @see WorkFlowTaskParameter. A UI layer can render
+ * appropriate components to collect user inputs for different scopes using these values
+ *
+ * @author Richard Wang (Github: RichardW98)
+ *
+ */
+public enum WorkFlowTaskParameterScope {
+
+	WORK_FLOW, TASK,
```
pkliczewski, line 27: What do you think about using `WORKFLOW`?

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/dto/converter/WorkFlowParametersConverter.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Workflow parameter converter
+ *
+ * @author Richard Wang (Github: richardw98)
+ * @author Annel Ketcha (Github: anludke)
+ */
+
+@Converter
+public class WorkFlowParametersConverter implements AttributeConverter<Set<WorkFlowTaskParameter>, String> {
```
pkliczewski, line 35: We support several parameter types including DATE and NUMBER. Do we want to provide way to convert them to corresponding types?

workflow-examples/src/main/java/com/redhat/parodos/examples/simple/LoggingWorkFlowTask.java
RichardW98: we used to have a feature allowing developer to define defaultValue for optional parameters.  I will try to restore it here

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/continuation/WorkFlowContinuationService.java
```diff
@@ -99,8 +102,12 @@ public void workFlowRunAfterStartup() {
 							throw new RuntimeException(e);
 						}
 					});
+					Map<String, String> workFlowArguments;
+					workFlowArguments = WorkFlowDTOUtil.readStringAsObject(workFlowExecution.getArguments(),
```
eloycoto, line 106: I would add this to the test, there is already one, so added the check will not hurt

tupyy:
stupid question: is saving the `user-id` compliant with GPDR?

anludke:
It's being reworked on master workflow refinement under feature/FLPATH-153

PR #74, opened 2023-02-28
Add unit tests coverage for NotificationRecordServiceImpl object

nirarg:
This change includes:
* Adding NotificationRecordServiceTests.java with all tests
* Minor code correctness changes in NotificationRecordServiceImpl.java
* Adding Java Docs to the interface - NotificationRecordService.java


notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationRecordServiceTests.java
```diff
@@ -0,0 +1,232 @@
...
+		exception = assertThrows(RuntimeException.class, () -> {
+			this.notificationRecordServiceImpl.getNotificationRecords(pageable, userName, State.UNREAD, searchTerm);
+		});
+		assertEquals("Search by state and search term combined not supported", exception.getMessage());
+	}
+
+	@Test
+	void countNotificationRecords() {
+		String userName = "test-user";
+		List<State> states = Arrays.asList(State.values());
```
gciavarrini, line 153: `Arrays.asList` is unnecessary here

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationRecordServiceTests.java
```diff
@@ -0,0 +1,232 @@
...
+					this.notificationRecordServiceImpl.countNotificationRecords(userName, state);
+				});
+				assertEquals(String.format("State %s is not supported", state), exception.getMessage());
+			}
+		}
+	}
+
+	@Test
+	void updateNotificationStatus() {
+		List<Operation> operations = Arrays.asList(Operation.values());
```
gciavarrini, line 185: `Arrays.asList` is unnecessary here

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationRecordServiceTests.java
```diff
@@ -0,0 +1,232 @@
...
+		this.notificationUserRepository = Mockito.mock(NotificationUserRepository.class);
+		this.notificationRecordServiceImpl = new NotificationRecordServiceImpl(notificationRecordRepository,
+				notificationUserRepository);
+	}
+
+	@Test
+	void createNotificationRecords() {
+		NotificationMessage notificationMessage = Mockito.mock(NotificationMessage.class);
+		NotificationUser notificationUser = Mockito.mock(NotificationUser.class);
+		List<NotificationUser> records = Arrays.asList(notificationUser);
```
gciavarrini, line 68: In such situation, I'd like to recommend to use `Collections.singletonList` instead of `Arrays.asList`. In this way, `notificationUser` will be Immutable. Since we're testing, this will increase the “control” on the object and on the array as well.

gciavarrini (commented):
In addition to my comments, I'd like to point out that there is a specific Jira task ([FLPATH-19](https://issues.redhat.com/browse/FLPATH-19)) for documenting interfaces, so maybe a separate PR can be more appropriate, but I don't have a strong opinion on this.

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationRecordServiceTests.java
nirarg: Removed this line and used the `State.values()` in the foreach loop

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationRecordServiceTests.java
nirarg: Removed this line and used the `Operation.values()` in the foreach loop

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationRecordServiceTests.java
nirarg: Thanks, changed all relevant places

nirarg:
> In addition to my comments, I'd like to point out that there is a specific Jira task ([FLPATH-19](https://issues.redhat.com/browse/FLPATH-19)) for documenting interfaces, so maybe a separate PR can be more appropriate, but I don't have a strong opinion on this.

Hi @gciavarrini, thank you for the review
I added here the Java doc only for the object that is tested, the java doc is tight to the unit tests (the unit tests must cover the "contract" described in the java doc)
I think this can be added in the same PR

gciavarrini (approved):
LGTM

lshannon (approved):

PR #75, opened 2023-02-28
FLPATH-43: Create and use custom Exception classes for throwables

gciavarrini:
This PR introduces custom exceptions that wrap uncheck Java exceptions.
Affected modules are: `workflow-service`, `workflow-engine`, and `notification-service`.

### DIFF
--- a/notification-service/src/main/java/com/redhat/parodos/notification/exceptions/StateNotFoundOrUnsupportedException.java
+++ b/notification-service/src/main/java/com/redhat/parodos/notification/exceptions/StateNotFoundOrUnsupportedException.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2022 Red Hat Developer
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


notification-service/src/main/java/com/redhat/parodos/notification/exceptions/StateNotFoundOrUnsupportedException.java
```diff
@@ -0,0 +1,44 @@
...
+
+	public StateNotFoundOrUnsupportedException() {
+	}
+
+	public StateNotFoundOrUnsupportedException(String message) {
+		super(message);
+	}
+
+	public StateNotFoundOrUnsupportedException(String message, Throwable cause) {
+		super(message, cause);
```
RichardW98, line 37: sonarqube analysis will fail if method is not referenced, can we remove this for now and add back when needed

RichardW98 (approved):

notification-service/src/main/java/com/redhat/parodos/notification/exceptions/StateNotFoundOrUnsupportedException.java
gciavarrini: Removed all unused methods from all the custom exceptions

RichardW98:
sqaush all commits into 1

PR #76, opened 2023-03-01
FLPATH-6: Add more tests coverage to notification-service

nirarg:


eloycoto (approved):

notification-service/src/test/java/com/redhat/parodos/notification/dto/NotificationAssemblerDTOTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+class NotificationAssemblerDTOTest {
+
+	@Test
+	void toModel() {
+		NotificationRecord entity = Mockito.mock(NotificationRecord.class);
+		NotificationMessage notificationMessage = Mockito.mock(NotificationMessage.class);
+
+		String messageBody = "message-body-test";
+		Instant messageCreation = Instant.now();
```
gciavarrini, line 24: To improve test quality, best practices recommend to do not use `Instant.now()` because you can't assert the exact value since it's different in each execution.
I understand that in this situation is not a critical point, but I'd like to suggest anyway to use something like `Instant.parse("2023-03-01T10:30:00.00Z")` to have a clean codebase.

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationMessageServiceTest.java
```diff
@@ -0,0 +1,59 @@
...
+		Mockito.when(notificationUserService.findUsers(usernames)).thenReturn(usersToNotify);
+
+		NotificationMessageService notificationMessageServiceImpl = new NotificationMessageServiceImpl(
+				notificationMessageRepository, notificationUserService, notificationRecordService, securityUtil);
+		notificationMessageServiceImpl.createNotificationMessage(notificationMessageCreateRequestDTO);
+
+		Mockito.verify(notificationMessageCreateRequestDTO, Mockito.times(1)).getMessageType();
+		Mockito.verify(notificationMessageCreateRequestDTO, Mockito.times(1)).getBody();
+		Mockito.verify(notificationMessageCreateRequestDTO, Mockito.times(1)).getSubject();
+		Mockito.verify(notificationMessageCreateRequestDTO, Mockito.times(2)).getUsernames();
```
gciavarrini, line 50: Question for learning: why are there two invocations of `getUsernames()` ?

notification-service/src/test/java/com/redhat/parodos/notification/dto/NotificationAssemblerDTOTest.java
nirarg: I use this in the beginning of the test so I can use the same Instant for all steps
In this case it doesn't matter what is the value assigned, the validation is done on the object itself, the same like when I generate random UUID
I can change it, but I don't think it must

notification-service/src/test/java/com/redhat/parodos/notification/service/NotificationMessageServiceTest.java
nirarg: The first time is used for building the Notification Message (inner function - `buildNotificationMessage`) and the second is used to find the Notification User (in the main called function - `createNotificationMessage`)

PR #82, opened 2023-03-06
AuthN: based all authentication on LDAP.

eloycoto:
This commit addresses authentication in the workflow-service using LDAP service. The behaviour is the following:

- For local development, it'll consume a local ldap unbounded service.
- For testing, a hacky ldap service is deployed, a super simple one and without many checkers.
- Local development profile is now extending values from Security auth profile, so things will not be messy in future changes.

The integration tests are now with CSRF&CORS enabled, so changes in the workflow examples are also made.

This commit address only Authentication; there is zero authorization in place.

Fix FLPATH-151


workflow-service/src/main/java/com/redhat/parodos/security/LocalSecurityConfiguration.java
```diff
@@ -36,43 +38,15 @@
 
 @Profile("local")
```
pkliczewski, line 39: We changed profile to `dev` [here](https://github.com/redhat-developer/parodos/pull/82/files#diff-a547c06eb8a140e1fc0f5e3b1d977f0055bf3ed5b09158296b24ff775e1fa878R39). Should this be updated as well?

workflow-service/src/main/java/com/redhat/parodos/security/SecurityConfiguration.java
```diff
@@ -33,16 +57,40 @@
  *
  */
 
+@Component
 @Configuration
 @Profile("!local")
```
pkliczewski, line 62: Same here

workflow-service/src/main/java/com/redhat/parodos/security/LocalSecurityConfiguration.java
eloycoto: It's ok, mainly because local uses the internal ldif server and dev uses the ldap connection to the openldap server.

workflow-service/src/main/java/com/redhat/parodos/security/SecurityConfiguration.java
eloycoto: https://github.com/redhat-developer/parodos/pull/82#discussion_r1128052180

workflow-service/src/main/java/com/redhat/parodos/WorkflowServiceApplication.java
```diff
@@ -30,6 +33,8 @@
 @EnableScheduling
 @EnableWebSecurity
 @EnableAspectJAutoProxy
+@EnableConfigurationProperties
+@ConfigurationPropertiesScan("com.redhat.parodos")
```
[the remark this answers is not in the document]
eloycoto: If I do not add that, It's not working :\

workflow-service/src/main/java/com/redhat/parodos/config/properties/LdapConnectionProperties.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.ConfigurationPropertiesScan;
+
+@ConfigurationProperties(prefix = "spring.ldap.connection")
+@ConfigurationPropertiesScan
+public class LdapConnectionProperties {
+
+	@Getter
+	@Setter
```
gciavarrini, line 13: Question: Do you use `@Setter` and `@Getter` instead of `@Data` because you want **both** getter and setter methods to be package private?
In addition, is there any need to support `toString` and `equals`?

workflow-service/src/main/resources/application-dev.yml
```diff
@@ -1,4 +1,17 @@
 logging:
   level:
     ROOT: INFO
-    com.redhat.parodos: DEBUG
\ No newline at end of file
+    com.redhat.parodos: INFO
```
gciavarrini, line 4: Why not debug for development?

workflow-service/src/main/java/com/redhat/parodos/config/properties/LdapConnectionProperties.java
eloycoto: toString or equals are not needed, the @Data I didn't know about it!

lshannon (approved):

PR #87, opened 2023-03-09
run_examples script: Fix run_simple_flow example

nirarg:
This includes the following changes:
* Get the PROJECT_ID and adding it the the Workflow execution
* Change parametes: name to workFlowName & tasks to workFlowTasks
* Add workflow task arguments to restCallTask
* Fix RestAPIWorkFlowTask use the right key for WorkFlowTaskParameters

### DIFF
--- a/workflow-examples/run_examples.sh
+++ b/workflow-examples/run_examples.sh
@@ -372,4 +372,16 @@ run_escalation_flow() {
   echo "                                                "
 }
 
-run_escalation_flow
+if [ $# -eq 0 ] || [ $1 = "escalation" ]; then
+  echo_blue "##### Running escalation flow #####"
+  run_escalation_flow
+elif [ $1 = "complex" ]; then
+  echo_blue "##### Running complex flow #####"
+  run_complex_flow
+elif [ $1 = "simple" ]; then
+echo_blue "##### Running simple flow #####"
+  run_simple_flow
+else
+  echo_red "##### Unsupported argument #####"
+  echo "Options: escalation (default) , complex, simple"
+fi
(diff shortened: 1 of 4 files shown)


workflow-examples/run_examples.sh
```diff
@@ -263,4 +310,5 @@ run_complex_flow() {
   echo "network workflow execution id:" $(echo_green $EXECUTION_ID)
 }
 
-run_complex_flow
+run_simple_flow
+# run_complex_flow
```
masayag, line 314: perhaps make this configurable with 'simple' as default?
e.g. For running the simple flow:
```
workflow-examples/run_examples.sh simple
# or
workflow-examples/run_examples.sh
```
and for complex:
```
workflow-examples/run_examples.sh complex
```
that way there won't be a need to modify the file when using it

nirarg: Added argument to control the flow type
Default type is complex, to keep previous behavior

workflow-examples/run_examples.sh
```diff
@@ -263,4 +310,13 @@ run_complex_flow() {
   echo "network workflow execution id:" $(echo_green $EXECUTION_ID)
 }
 
-run_complex_flow
+if [ $# -eq 0 ] || [ $1 = "complex" ]; then
```
pkliczewski, line 313: Do you think it makes sense to add support to run both types sequentially?

eloycoto: please, do not over complicated when yesterday we agreed to moved this script to another things :-)

nirarg: I agree with @eloycoto, I prefer to keep it like this and open tickets for new requirements

pkliczewski:
/lgtm

Please wait with the merge. I would like to merge this [PR](https://github.com/redhat-developer/parodos/pull/93) first.

lshannon (approved):

PR #90, opened 2023-03-10
Tools: add a way to install sample backstage in plain kubernetes

eloycoto:


hack/manifests/backstage/Readme.md
```diff
@@ -0,0 +1,56 @@
+# Install parados on plain kubernetes
+
+
+## Install kind
```
gabriel-farache, line 4: this title is confusing: I expected steps to install `kind`  (or a link to https://kind.sigs.k8s.io/docs/user/quick-start/) but instead, it is the command to initialize the local cluster with kind

hack/manifests/backstage/Readme.md
```diff
@@ -0,0 +1,55 @@
+# Install parados on plain kubernetes
```
pkliczewski, line 1: what about having make target for this?

eloycoto: It changed a bit, it was a draft PR, look at it now. I do not think that makes sense, but yeah.

hack/manifests/backstage/postgres.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: apps/v1
+kind: Deployment
```
RichardW98, line 3: my understanding is that most of the time, db should be statefulset for persist storage.  in the future our backend need to use an external db too for production(h2 should only be for local testing)

hack/manifests/backstage/app-config.yaml
```diff
@@ -0,0 +1,116 @@
...
+    # Default Helmet Content-Security-Policy values can be removed by setting the key to false
+  cors:
+    origin: http://localhost:8080
+    methods: [GET, HEAD, PATCH, POST, PUT, DELETE]
+    credentials: true
+
+  database:
+    client: pg
+    connection:
+      host: ${POSTGRES_SERVICE_HOST}
```
RichardW98, line 32: postgres.default.svc.cluster.local

eloycoto: nope, that means the namespace is the default, using this way will inherit the right one.

hack/manifests/backstage/postgres.yaml
RichardW98: yes let's keep it in deploy

RichardW98 (approved):

PR #92, opened 2023-03-10
Enforce java version in Makefile

gabriel-farache:
As we had issue with tests caused by wrong java version (both java and java version of mvn), checks are now enforced in the Makefile to make sure the java version set is within supported bounds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ java-checks:
     $(error "No java found in $(PATH). Please install java >= 11")
   else
     JAVA_VERSION=$(shell java --version | head -n 1 | sed -rn "s/.*\s([[:digit:]]+)\.[[:digit:]]+\.[[:digit:]]+\s.*/\1/p")
-    JAVA_VERSION_IS_SUPPORTED=$(shell [ $(JAVA_VERSION) -le $(JAVA_VERSION_MAX_SUPPORTED)  ] && [ $(JAVA_VERSION) -ge $(JAVA_VERSION_MIN_SUPPORTED)  ]&& echo true)
+    JAVA_VERSION_IS_SUPPORTED=$(shell [ $(JAVA_VERSION) -le $(JAVA_VERSION_MAX_SUPPORTED)  ] && [ $(JAVA_VERSION) -ge $(JAVA_VERSION_MIN_SUPPORTED)  ] && echo true)
     ifeq ($(JAVA_VERSION_IS_SUPPORTED), )
       $(error "Java version $(JAVA_VERSION) should be [$(JAVA_VERSION_MIN_SUPPORTED) ; $(JAVA_VERSION_MAX_SUPPORTED)]. Please install Java within those bounds")
     endif
--- a/Makefile
+++ b/Makefile
@@ -25,27 +25,29 @@ JAVA_VERSION_MIN_SUPPORTED=11
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
+mvn-checks:
+  # check if maven is installed
+  ifeq (,$(shell which $(MAVEN)))
+    $(error "No maven found in $(PATH). Please install maven")
+  else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,16 +29,23 @@ SHELL = /usr/bin/env bash -o pipefail
 # check if maven is installed
 ifeq (,$(shell which $(MAVEN)))
   $(error "No maven found in $(PATH). Please install maven")
+else
+  MVN_JAVA_VERSION=$(shell mvn --version | egrep -o "Java version: [[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+"| head -n 1 | cut -c15-16)
```
tupyy, line 33: If you want to enforce why not looking directly for what you want?
suggestion:
```
mvn --version | egrep ":\s*($JAVA_VERSION_MAX_SUPPORTED|$JAVA_VERSION_MIN_SUPPORTED)\.[[:digit:]]+\.[[:digit:]]+"
```

masayag (approved):

masayag (changes requested):
taken the approval off - noticed this is being examined for each target, even for targets that builds in a container with the proper java version:
```
[containers](https://github.com/redhat-developer/parodos/actions/runs/4387066775/jobs/7681971939#logs)
failed yesterday in 9s
Search logs
1s
1s
3s
2s
Run make build-images
  make build-images
  shell: /usr/bin/bash -e {0}
Makefile:36: *** "Maven Java [1](https://github.com/redhat-developer/parodos/actions/runs/4387066775/jobs/7681971939#step:4:1)1 version should be [11 ; 11]. Please use Java within these bounds with mvn".  Stop.
Error: Process completed with exit code [2](https://github.com/redhat-developer/parodos/actions/runs/4387066775/jobs/7681971939#step:4:2).
```

Makefile
```diff
@@ -27,16 +29,23 @@ SHELL = /usr/bin/env bash -o pipefail
...
+  MVN_VERSION_IS_SUPPORTED=$(shell [ $(MVN_JAVA_VERSION) -lt $(JAVA_VERSION_MAX_SUPPORTED)  ] && [ $(MVN_JAVA_VERSION) -gt $(JAVA_VERSION_MIN_SUPPORTED)  ] && echo true)
```
tupyy, line 34: shouldn't be `le` and `ge` ?

Makefile
tupyy, line 33: just a suggestion. I would use `sed` instead of `cut`. it is too rigit.
```
mvn --version | sed -rn 's/.*([[:digit:]]{2})\.[[:digit:]]+\.[[:digit:]]+(,|\s).*/\1/p'
```

Makefile
gabriel-farache: yes

pkliczewski (commented):
/lgtm

Makefile
```diff
@@ -18,26 +18,36 @@ VERSION = $(shell sed -n "s/<revision>\(.*\)<\/revision>/\1/p" $(PWD)/pom.xml |
...
+    endif
+  endif
+
+java-checks:
+  # check java version
+  ifeq (,$(shell java --version))
+    $(error "No java found in $(PATH). Please install java >= 11")
+  else
+    JAVA_VERSION=$(shell java --version | head -n 1 | sed -rn "s/.*\s([[:digit:]]+)\.[[:digit:]]+\.[[:digit:]]+\s.*/\1/p")
+    JAVA_VERSION_IS_SUPPORTED=$(shell [ $(JAVA_VERSION) -le $(JAVA_VERSION_MAX_SUPPORTED)  ] && [ $(JAVA_VERSION) -ge $(JAVA_VERSION_MIN_SUPPORTED)  ]&& echo true)
```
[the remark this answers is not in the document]
tupyy: it does not work with true because if ifeq. you need something to check agaist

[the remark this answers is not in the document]
gabriel-farache: > bash is not yelling because of ]]&& echo true at the end? generally is very picky about this stuff
Nope not yelling, I guess it's because it's enclosed in `shell` but I'll add the space

> btw, why echo true and not just true
As @tupyy said, if I don't echo anything, nothing will be stored in the variable tested in `ifreq`, in plain bash I would test `$?`

lshannon (approved):

PR #94, opened 2023-03-12
Extract getBase64Creds to a shared module

masayag:
In order to simplify project dependencies, the explicit dependency from `workflow-service` module to `workflow-example` is being reduced by extracting to a shared module the required functionality.

### DIFF
--- a/parodos-model-api/src/main/java/com/redhat/parodos/workflow/utils/CredUtils.java
+++ b/parodos-model-api/src/main/java/com/redhat/parodos/workflow/utils/CredUtils.java
@@ -1,0 +1,21 @@
+package com.redhat.parodos.workflow.utils;
+
+import java.util.Base64;
+
+public class CredUtils {
+
+	/**
+	 * Generates a Base64 encoding of username:password string
+	 * @param username the username
+	 * @param password the password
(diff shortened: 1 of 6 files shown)


workflow-engine/src/main/java/com/redhat/parodos/workflows/utils/CredUtils.java
```diff
@@ -0,0 +1,21 @@
+package com.redhat.parodos.workflows.utils;
+
+import java.util.Base64;
+
+public class CredUtils {
```
RichardW98, line 5: please move this class to parodos-model-api. Don't add any non-workflow-related code in workfow-engine. Btw,  workflow-engine is based on an existing public project: https://github.com/j-easy/easy-flows. Please let's be careful when adding code there. Any base code soley for Parodos should be in parodos-model-api. @pkliczewski @eloycoto @lshannon @anludke @gciavarrini please advise.

masayag: Thanks for clarifying the differentiation between the engine and model-api.
I moved the class to model-api.

gciavarrini (approved):
lgtm

PR #97, opened 2023-03-13
fix workflow name

RichardW98:
fix workflow name
workflow name should be consistent with its bean name for aspect discovery

### DIFF
--- a/workflow-examples/src/main/java/com/redhat/parodos/examples/continued/complex/ComplexWorkFlowConfiguration.java
+++ b/workflow-examples/src/main/java/com/redhat/parodos/examples/continued/complex/ComplexWorkFlowConfiguration.java
@@ -66,7 +66,7 @@ public class ComplexWorkFlowConfiguration {
 			@Qualifier("onboardingAssessmentTask") OnboardingAssessmentTask onboardingAssessmentTask) {
 		// @formatter:off
 		return SequentialFlow.Builder.aNewSequentialFlow()
-				.named("onboarding Assessment WorkFlow")
+				.named("onboardingAssessment" + WorkFlowConstants.ASSESSMENT_WORKFLOW)
 				.execute(onboardingAssessmentTask)
 				.build();
 		// @formatter:on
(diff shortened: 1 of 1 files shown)


workflow-examples/src/main/java/com/redhat/parodos/examples/continued/complex/ComplexWorkFlowConfiguration.java
```diff
@@ -66,7 +66,7 @@ WorkFlow assessmentWorkFlow(
 			@Qualifier("onboardingAssessmentTask") OnboardingAssessmentTask onboardingAssessmentTask) {
 		// @formatter:off
 		return SequentialFlow.Builder.aNewSequentialFlow()
-				.named("onboarding Assessment WorkFlow")
+				.named("onboardingAssessment" + WorkFlowConstants.ASSESSMENT_WORKFLOW)
```
masayag, line 69: A question about this one: Isn't it redundant to specify the workflow type by annotation and also by the name?

RichardW98: good question Moti! its name is just a metadata and it should be consistent from bean to builder(basically you can give a workflow any name). Annotation is which type it really is, we need to process a workflow based on its annotated type(e.g. assessment should return options,  checker should run on a schedule for certain tasks and so on).

RichardW98: let me know if it makes sense

gciavarrini (approved):
lgtm

workflow-examples/src/main/java/com/redhat/parodos/examples/continued/complex/ComplexWorkFlowConfiguration.java
anludke: @masayag Workflow type in parados is either Infrastructure, Assessment and Checker. Annotations haven't introduced to make that distinction including the need for additional metadata. On the other hand, workflow processing type (Sequential or Parallel) is derived form the workflow bean. Hope that helps clarifying the workflow type part.

masayag: > Workflow type in parados is either Infrastructure, Assessment and Checker. Annotations

and Escalation since the weekend ;-) 

Thank you both for your replies. That makes sense to me.

PR #99, opened 2023-03-13
CI: Fix workflow-example

eloycoto:
Fix workflow-example CI test and also fail if any command fails

### DIFF
--- a/workflow-examples/run_examples.sh
+++ b/workflow-examples/run_examples.sh
@@ -5,6 +5,9 @@
 #process_response() {
 #  echo "$1" | awk -F\" '{print $2}'
 #}
+#
+
+set -e
 
 SERVERIP=${SERVERIP:-127.0.0.1}
 SERVERPORT=${SERVERPORT:-8080}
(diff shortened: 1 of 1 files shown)


workflow-examples/run_examples.sh
```diff
@@ -89,6 +92,24 @@ get_next_workflow() {
     -H 'Content-Type: application/json' | jq '.tasks[] | select(has("nextWorkFlow"))' | jq -r '.nextWorkFlow' | head -n 1
 }
 
+wait_project_start() {
+  echo_blue "******** Checking project is running ********"
+  for i in {1..1}
```
masayag, line 97: `1..1` is a single iteration, right?

masayag: meaning, `i` isn't progressing, so only one iteration and the condition on 108 will always be false?

eloycoto: sorry, was testing it and I forget it.

RichardW98 (approved):

pkliczewski (commented):
/lgtm

PR #100, opened 2023-03-13
WorkFlowExecutionAspect: Avoid NullPointerException on execution failure

nirarg:
In case the execution failed and the WorkReport is null, create DefaultWorkReport With WorkStatus=FAILED, in order to prevent NullPointerException


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/aspect/WorkFlowExecutionAspect.java
```diff
@@ -115,7 +113,8 @@ public WorkReport executeAroundAdvice(ProceedingJoinPoint proceedingJoinPoint, W
 			report = (WorkReport) proceedingJoinPoint.proceed();
 		}
 		catch (Throwable e) {
-			log.error("Workflow {} has failed!", workFlowName);
+			report = new DefaultWorkReport(WorkStatus.FAILED, workContext);
+			log.error("Workflow {} has failed! with error: {}", workFlowName, e);
```
eloycoto, line 117: maybe adding the unit test will be a good idea!

anludke (approved):

PR #107, opened 2023-03-14
Add 'provided' scope to existing jars in examples

masayag:
Add 'provided' scope to existing jars in examples
    
In order to rely on jar versions as provided by the workflow server, there is a need to set the scope of those dependencies in workflow-examples/pom.xml as `provided`.
    
Signed-off-by: Moti Asayag <masayag@redhat.com>


pkliczewski (commented):
/lgtm

RichardW98 (approved):

PR #110, opened 2023-03-14
Flpath 189: Add WorkStatus to WorkFlowResponseDTO and set it before return to user

nirarg:


workflow-examples/run_examples.sh
```diff
@@ -109,6 +109,31 @@ wait_project_start() {
...
+    -H 'Authorization: Basic dGVzdDp0ZXN0' \
+    -H "X-XSRF-TOKEN: ${TOKEN}" \
+    -b $COOKIEFP \
+    -H 'Content-Type: application/json' \
+    -d "$params")
+
+    status=$(echo "$response" | jq -r '.workStatus')
+    if [[ "$status" == "FAILED" ]]; then
+      if "$exitOnFailure"; then
+        echo_red "Error: HTTP request failed" >&2
```
eloycoto, line 127: I would use @fail function here tbh

nirarg: Done

PR #113, opened 2023-03-15
WIP: Shell runner tasks

rgolangh:
Introduce basic shell tasks that consumes {"command": "", "args": ""}
executes, and timeouts after a default of 10 minutes.
Pass "timeout": [timeout in seconds] to override


pkliczewski (commented):
code changes looks good
@ydayagi please review. It seems you wanted to place tasks in different [place](https://github.com/ydayagi/parodos/commit/4d85eeac01b72113814998d29964a0cb9e27a12e)

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTask.java
```diff
@@ -0,0 +1,94 @@
...
+		ShellRunnerTaskParams(WorkContext workContext) {
+			String command = (String) workContext.get("command");
+			if (command == null || command.isBlank()) {
+				throw new IllegalArgumentException("argument 'command' is empty or blank");
+			}
+			this.cmdline = new ArrayList<>() {
+			};
+			this.cmdline.add(command);
+			if (workContext.get("args") != null) {
+				this.cmdline.addAll(List.of(workContext.get("args").toString().split(" ")));
```
eloycoto, line 49: Is this not a security issue? A user can submit a command that scales privilege, no? like sh - sudo rm

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTask.java
```diff
@@ -0,0 +1,94 @@
...
+				r.lines().forEach(outputConsumer);
+			}
+			log.info("ended shell task invocation");
+			return new DefaultWorkReport(p.exitValue() == 0 ? WorkStatus.COMPLETED : WorkStatus.FAILED, workContext);
+		}
+		catch (IOException | InterruptedException e) {
+			throw new RuntimeException(e);
+		}
+		finally {
+			tmpDir.delete();
```
eloycoto, line 90: If you return in the try-and-catch, this is never going to be executed, no?

parodos-model-api/src/test/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTaskTest.java
```diff
@@ -0,0 +1,63 @@
...
+	@Test
+	public void executeWithOutput() {
+		UUID randomUUID = UUID.randomUUID();
+		ctx.put("command", "echo");
+		ctx.put("args", randomUUID);
+		var output = new StringBuilder();
+		underTest.outputConsumer = output::append;
+
+		assertThat(underTest.execute(ctx).getStatus()).isEqualTo(WorkStatus.COMPLETED);
+		assertThat(output.toString()).isEqualTo(randomUUID.toString());
```
eloycoto, line 34: could we assert that the temp folder is deleted?

eloycoto (commented):
A few minor comments, thanks for the work!

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTask.java
tupyy: finally is always executed

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTask.java
rgolangh: very very good point. The think is that I really don't know what the deployment model is yet. For example, will the user the jvm runs under is in `sudoers` at all? could we actually ensure podman is part of the deployment and so run every execution part of a ubi9/micro image?

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTask.java
rgolangh: that is exactly what finally is for. It is always executed, unless there's a `System.exit` statement

parodos-model-api/src/test/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTaskTest.java
rgolangh: following the comment above, we can count on the jvm to execute that. but for extra cation we can parametrize the call to pass in a temp dir ... wonder if its worth it.
Another though I had - writing  stuff into /tmp costs memory while /var/tmp is disk space only and I tend to think those executions could involve a lot of stuff written into the scratch space of the task .

nirarg:
/hold
(test label)

parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/shell/ShellRunnerTask.java
eloycoto: ok, and also, I would run at least in a separate shell, like sh 'command', if not can get the ENV variables and maybe connect to the database itself. 

TBH, I would move this to "container shell command", and be able to specify the image and the command to execute, for me, it will be 10000 times safer.

pkliczewski: We still need to work on deployment story and we can make sure that ability to exploit this issue is limited or not possible. Let's discuss this issue while working on FLPATH-198.

lshannon:
The lgtm label should have been applied. Not sure why its still saying its not. Going to go ahead and merge this

nirarg:
@rgolangh This code should be located in `prebuilt-tsks` module, no?

PR #116, opened 2023-03-15
remove old complex workflow and replace it with master Workflow

RichardW98:
the old complex workflow uses our old logic with next workflow,  we need to change to use the new master workflow.

Screenshot 2023-03-15 at 11 53 32 AM


workflow-examples/src/main/java/com/redhat/parodos/examples/simple/task/LoggingWorkFlowTask.java
```diff
@@ -57,7 +57,7 @@ public WorkReport execute(WorkContext workContext) {
 		catch (Exception e) {
 			log.error("There was an issue with the task {}: {}", getName(), e.getMessage());
 		}
-		return new DefaultWorkReport(WorkStatus.COMPLETED, workContext);
+		return new DefaultWorkReport(WorkStatus.FAILED, workContext);
```
gciavarrini, line 60: Why does the LoggingWorkflowTask need to fail always?

workflow-examples/src/test/java/com/redhat/parodos/examples/master/checker/NamespaceApprovalWorkFlowCheckerTaskTest.java
```diff
@@ -24,19 +23,19 @@
  *
  * @author Gloria Ciavarrini (Github: gciavarrini)
  */
-public class MockApprovalWorkFlowCheckerTaskTest extends BaseWorkFlowCheckerTaskTest {
+public class NamespaceApprovalWorkFlowCheckerTaskTest extends BaseWorkFlowCheckerTaskTest {
```
gciavarrini, line 26: Can you update the description and the author as well?

workflow-examples/src/main/java/com/redhat/parodos/examples/master/MasterWorkFlowConfiguration.java
```diff
@@ -24,6 +28,35 @@
 @Configuration
 public class MasterWorkFlowConfiguration {
 
+	// Assessment workflow
+	@Bean
+	WorkFlowOption onboardingOption() {
+		return new WorkFlowOption.Builder("onboardingOption", "masterWorkFlow")
+				.addToDetails("An example of a complex WorkFlow with Status checks").displayName("Onboarding")
```
gciavarrini, line 35: nit: I think that `displayName` should be in a new line

workflow-examples/src/main/java/com/redhat/parodos/examples/simple/task/LoggingWorkFlowTask.java
RichardW98: it should fail after exception is caught

gciavarrini: Oh right! Thank you for the clarification!

workflow-examples/src/main/java/com/redhat/parodos/examples/master/MasterWorkFlowConfiguration.java
RichardW98: the spring formatter forced to keep it in the that line

pkliczewski (commented):
/lgtm

@gciavarrini please take a look

PR #118, opened 2023-03-16
Manifest: Add health endpoints

eloycoto:
To make sure that the service is ready before marking it in Kubernetes and starting to listen traffict to.


pkliczewski (commented):
/lgtm

PR #119, opened 2023-03-16
Generate SDK from OpenAPI JSON

gciavarrini:
This PR adds a new module `workflow-service-sdk`.

Using the open API JSON file, via OpenAPI Tool maven generator, the workflow service SDK is generated.
Description on how to generate the SDK is available on [workflow-service-sdk/README.md](https://github.com/redhat-developer/parodos/pull/119/files#diff-b9283fc27f1d0e6a641c28e85bfb2da4276c12f3fbbafcc367c567c4321cb7b6)

This PR adds also a minimal example on how to use the SDK in `workflow-example` module.


pkliczewski (commented):
It would be useful to add instruction how to regenerate schema to the README.

gciavarrini:
> It would be useful to add instruction how to regenerate schema to the README.
Thanks for the suggestion!

I added a generated README. WDYT?

workflow-service-sdk/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+                <artifactId>download-maven-plugin</artifactId>
+                <version>1.6.8</version>
+                <executions>
+                    <execution>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>wget</goal>
+                        </goals>
+                        <configuration>
+                            <url>http://localhost:8080/v3/api-docs</url>
```
gciavarrini, line 74: @pkliczewski @eloycoto 
Any suggestion on how can I get this working?

I'm able to download the JSON file in my local env but it's not working here.

workflow-service-sdk/.gitignore
```diff
@@ -0,0 +1,21 @@
+*.class
```
eloycoto, line 1: this file should be from the previous folder, no?

workflow-service-sdk/pom.xml
eloycoto: what access this port? I'm not sure If I understand the flow.

.gitignore
```diff
@@ -60,4 +60,18 @@ local.properties
...
 ## VSCODE ***
-/.vscode/
\ No newline at end of file
+/.vscode/
+
+##############################
+## Workflow Service SDK Generated files
+##############################
+
+workflow-service-sdk/src/
```
masayag, line 69: I don't think we want this :-) 
It will ignore any changes to the project.
Perhaps there is an option to specify the output directory for the generated resource?

workflow-service-sdk/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Maven users
+
+Add this dependency to your project's POM:
+
+```xml
+<dependency>
+  <groupId>dev.parodos</groupId>
+  <artifactId>workflow-service-sdk</artifactId>
+  <version>v1.0.0-SNAPSHOT</version>
```
masayag, line 43: version schema should match the existing code-base which is 1.0.2-SNAPSHOT.
Is this file auto-generated?

workflow-service-sdk/README.md
```diff
@@ -0,0 +1,155 @@
...
+*ProjectApi* | [**getProjects**](docs/ProjectApi.md#getProjects) | **GET** /api/v1/projects | Returns a list of project
+*WorkflowApi* | [**execute**](docs/WorkflowApi.md#execute) | **POST** /api/v1/workflows | Executes a workflow
+*WorkflowApi* | [**getStatus**](docs/WorkflowApi.md#getStatus) | **GET** /api/v1/workflows/{id}/status | Returns a workflow status
+*WorkflowDefinitionApi* | [**getWorkFlowDefinitionById**](docs/WorkflowDefinitionApi.md#getWorkFlowDefinitionById) | **GET** /api/v1/workflowdefinitions/{id} | Returns information about a workflow definition by id
+*WorkflowDefinitionApi* | [**getWorkFlowDefinitions**](docs/WorkflowDefinitionApi.md#getWorkFlowDefinitions) | **GET** /api/v1/workflowdefinitions | Returns a list of workflow definition
+
+
+## Documentation for Models
+
+ - [ArgumentRequestDTOGenerated](docs/ArgumentRequestDTOGenerated.md)
```
masayag, line 128: Is there an option to get rid of the suffix `DTOGenerated` ?

workflow-service-sdk/pom.xml
masayag: IIUC, for this to work, there is a need to have the workflow-service running.
In https://github.com/springdoc/springdoc-openapi-maven-plugin the expectation is to run this during the integration-test to make sure the service is up and accessible.
Perhaps there is an option to run a goal to start the application for the `/v3/api-docs` to be available, generate the resource and then stop the app.

workflow-service-sdk/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+                            <artifactId>workflow-service-sdk</artifactId>
+                            <packageName>com.redhat.parodos.openapi</packageName>
+
+                            <inputSpec>${project.basedir}/api-docs.json</inputSpec>
+                            <output>${basedir}</output>
+
+                            <invokerPackage>com.redhat.parodos.openapi.api</invokerPackage>
+                            <apiPackage>com.redhat.parodos.openapi.api</apiPackage>
+                            <modelPackage>com.redhat.parodos.openapi.model</modelPackage>
+                            <modelNameSuffix>Generated</modelNameSuffix>
```
masayag, line 104: I don't think this is needed. The name of the package should highlight its origin.
We can change the modelPackage to `com.redhat.parodos.sdk.model` so it will be clear this isn't the model.

workflow-service-sdk/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.openapitools</groupId>
+                <artifactId>openapi-generator-maven-plugin</artifactId>
+                <version>${openapi.generator.version}</version>
+                <executions>
+                    <execution>
+                        <id>generate-client-api-code</id>
+                        <goals>
+                            <goal>generate</goal>
```
masayag, line 89: Shouldn't the generated code be added to this PR?

.gitignore
gciavarrini: I'm already using the [option](https://github.com/redhat-developer/parodos/blob/de8579b246d6991d08f63621cc41db09a266f298/workflow-service-sdk/pom.xml#L99) to specify the output directory. 

So my doubt here is: do we want to commit/push any change in the generated code? I thought we are interested in the library jar, that's why I added this to the `.gitignore`.

workflow-service-sdk/README.md
gciavarrini: I added the suffix `Generated` [here](https://github.com/redhat-developer/parodos/blob/de8579b246d6991d08f63621cc41db09a266f298/workflow-service-sdk/pom.xml#L104) I'll remove it :-)  

The `DTO` suffix is created by the tool, I'm not able to remove it.

workflow-service-sdk/README.md
gciavarrini: Yes, It's auto generated.
Fixed!

workflow-service-sdk/.gitignore
gciavarrini: Do you mean that there shouldn't be a `gitignore` file in this module? 
Or you don't like something inside the `.gitignore` file (ie. `*.class`)? 
This `.gitignore` file is auto generated by the openapi tool.
TBH, I've no idea on how to customize its generation. The only thing that I can do is to add the `.gitignore` file to `.openapi-generator-ignore` to prevent the generation and the manually "tune" the `.gitignore`.

Can you clarify a bit?

workflow-service-sdk/pom.xml
pkliczewski: Is there any other way we can get the schema? Running the service may not be that easy.

eloycoto: can we commit the openapi schema to a file, and keep it sync in git?

workflow-service-sdk/.gitignore
eloycoto: ok, make sense.

workflow-service-sdk/pom.xml
rgolangh: yes the server needs to boot and then it read all the annotations on the classes comprising the api , and then generate the schema. It makes sense to generate this and keep it tied to a version. 
So the ci would have to test and make sure the resulting openapi.json doesn't have changes. Developer must make sure the generate ran (don't remeber which mvn target this hooks into. could be a make target to make it simpe)

eloycoto: +1 here, we can use something like we had in flotta docs:
https://github.com/project-flotta/project-flotta.github.io/blob/308f37345174dde14be987c901590ff030f3968d/.github/workflows/crd_update.yaml#L24

.gitignore
```diff
@@ -60,4 +60,18 @@ local.properties
...
+workflow-service-sdk/api/
+workflow-service-sdk/docs/
+workflow-service-sdk/*gradle*
+workflow-service-sdk/.openapi-generator/
+workflow-service-sdk/build.*
+workflow-service-sdk/api-docs.json
+workflow-service-sdk/git_push.sh
+workflow-service-sdk/.travis.yml
```
nirarg, line 77: Shouldn't those lines be added to the `.gitignore` file inside the module, if it already exists?

nirarg (commented):
The OpenAPI SDK should be used also for the `notification-service`
I don't think it should be added in this PR.
But I think we need to consider the following:
Should we have different module for each service sdk, or all services sdk files can be added in the same module?
If the second option is used, need to have more generic name for the module

pom.xml
```diff
@@ -62,6 +62,7 @@
         <module>notification-service</module>
         <module>pattern-detection-library</module>
         <module>coverage</module>
+        <module>workflow-service-sdk</module>
```
pkliczewski, line 65: It would be great to add one more module for notification service sdk.

nirarg: Open new Jira task for this https://issues.redhat.com/browse/FLPATH-208

PR #123, opened 2023-03-17
Add a workflow task that sends a message to topic in TIBCO

ydayagi:
[FLPATH-145](https://issues.redhat.com//browse/FLPATH-145) https://issues.redhat.com/browse/FLPATH-145

### DIFF
--- a/external-dependencies/download-tibco.sh
+++ b/external-dependencies/download-tibco.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+if [ -f "tibjms.jar" ]; then exit 0; fi
+rm -rf /tmp/tibco
+mkdir /tmp/tibco
+curl -s -o /tmp/tibco/tibco.zip "https://edownloads.tibco.com/Installers/tap/EMS-CE/10.2.1/TIB_ems-ce_10.2.1_linux_x86_64.zip?SJCDPTPG=1681227113_1e3fe1f60cbbe9ee810af68e0b4256bb&ext=.zip"
+cd /tmp/tibco
+unzip tibco.zip TIB_ems-ce_10.2.1/tar/TIB_ems-ce_10.2.1_linux_x86_64-java_client.tar.gz
+tar -zxf TIB_ems-ce_10.2.1/tar/TIB_ems-ce_10.2.1_linux_x86_64-java_client.tar.gz  opt/tibco/ems/10.2/lib/tibjms.jar
+cd -
+cp /tmp/tibco/opt/tibco/ems/10.2/lib/tibjms.jar .
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,8 @@
         <module>notification-service</module>
         <module>pattern-detection-library</module>
         <module>coverage</module>
+        <module>external-dependencies</module>
+        <module>prebuilt-tasks</module>
     </modules>
     <build>
         <pluginManagement>
--- a/prebuilt-tasks/pom.xml
+++ b/prebuilt-tasks/pom.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.parodos</groupId>
+        <artifactId>parados-parent</artifactId>
+        <version>${revision}</version>
+    </parent>
+
+    <artifactId>prebuilt-tasks</artifactId>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>dev.parodos</groupId>
+            <artifactId>external-dependencies</artifactId>
+            <version>${revision}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.24</version>
+        </dependency>
+        <dependency>
+            <groupId>dev.parodos</groupId>
+            <artifactId>parodos-model-api</artifactId>
+            <version>${revision}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.jms</groupId>
+            <artifactId>jakarta.jms-api</artifactId>
+            <version>2.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>com.tibco</groupId>
+            <artifactId>tibjms</artifactId>
+            <version>10.2</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.13.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.9.0</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTask.java
+++ b/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTask.java
@@ -1,0 +1,74 @@
+package com.redhat.parodos.tasks.tibco;
+
+import com.redhat.parodos.workflow.task.BaseWorkFlowTask;
+import com.redhat.parodos.workflow.task.WorkFlowTaskOutput;
+import com.redhat.parodos.workflow.task.parameter.WorkFlowTaskParameter;
+import com.redhat.parodos.workflow.task.parameter.WorkFlowTaskParameterType;
+import com.redhat.parodos.workflows.work.DefaultWorkReport;
+import com.redhat.parodos.workflows.work.WorkContext;
+import com.redhat.parodos.workflows.work.WorkReport;
+import com.redhat.parodos.workflows.work.WorkStatus;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.LinkedList;
+import java.util.List;
+
+@Slf4j
+public class TibcoWorkFlowTask extends BaseWorkFlowTask {
+
+	private final Tibjms service;
+
+	private final String url;
+
+	private final String caFile;
+
+	private final String username;
+
+	private final String password;
+
+	public TibcoWorkFlowTask(String url, String caFile, String username, String password) {
+		this(new TibjmsImpl(), url, username, password, caFile);
+	}
+
+	TibcoWorkFlowTask(Tibjms service, String url, String caFile, String username, String password) {
+		this.service = service;
+		this.url = url;
+		this.caFile = caFile;
+		this.username = username;
+		this.password = password;
+	}
+
+	@Override
+	public @NonNull List<WorkFlowTaskParameter> getWorkFlowTaskParameters() {
+		LinkedList<WorkFlowTaskParameter> params = new LinkedList<WorkFlowTaskParameter>();
+		params.add(WorkFlowTaskParameter.builder().key("topic").type(WorkFlowTaskParameterType.TEXT).optional(false)
+				.description("Topic to send to").build());
+		params.add(WorkFlowTaskParameter.builder().key("message").type(WorkFlowTaskParameterType.TEXT).optional(false)
+				.description("Message to send to topic").build());
+		return params;
+	}
+
+	@Override
+	public @NonNull List<WorkFlowTaskOutput> getWorkFlowTaskOutputs() {
+		return List.of(WorkFlowTaskOutput.OTHER);
+	}
+
+	@Override
+	public WorkReport execute(WorkContext workContext) {
+		try {
+			String topic = getParameterValue(workContext, "topic");
+			String message = getParameterValue(workContext, "message");
+
+			service.sendMessage(url, caFile, username, password, topic, message);
+
+		}
+		catch (Exception e) {
+			log.error("TIBCO task failed", e);
+			return new DefaultWorkReport(WorkStatus.FAILED, workContext, e);
+		}
+
+		return new DefaultWorkReport(WorkStatus.COMPLETED, workContext);
+	}
+
+}
--- a/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/Tibjms.java
+++ b/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/Tibjms.java
@@ -1,0 +1,8 @@
+package com.redhat.parodos.tasks.tibco;
+
+public interface Tibjms {
+
+	public void sendMessage(String url, String caFile, String username, String password, String topic, String message)
+			throws javax.jms.JMSException;
+
+}
--- a/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
+++ b/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
@@ -1,0 +1,30 @@
+package com.redhat.parodos.tasks.tibco;
+
+import com.tibco.tibjms.TibjmsConnectionFactory;
+
+import javax.jms.JMSException;
+import javax.jms.Connection;
+import javax.jms.Destination;
+import javax.jms.Session;
+import javax.jms.MessageProducer;
+import javax.jms.TextMessage;
+
+public class TibjmsImpl implements Tibjms {
+
+	@Override
+	public void sendMessage(String url, String caFile, String username, String password, String topic, String message)
+			throws JMSException {
+		TibjmsConnectionFactory factory = new TibjmsConnectionFactory(url);
+		if (!caFile.isEmpty()) {
+			factory.setSSLTrustedCertificate(caFile);
+		}
+		try (Connection connection = factory.createConnection(username, password);
+				Session session = connection.createSession(javax.jms.Session.AUTO_ACKNOWLEDGE);) {
+			Destination destination = session.createTopic(topic);
+			MessageProducer producer = session.createProducer(destination);
+			TextMessage textMessage = session.createTextMessage(message);
+			producer.send(textMessage);
+		}
+	}
+
+}
--- a/prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
+++ b/prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
@@ -1,0 +1,81 @@
+package com.redhat.parodos.tasks.tibco;
+
+import com.redhat.parodos.workflow.context.WorkContextDelegate;
+import com.redhat.parodos.workflow.exception.MissingParameterException;
+import com.redhat.parodos.workflows.work.WorkContext;
+import com.redhat.parodos.workflows.work.WorkReport;
+import com.redhat.parodos.workflows.work.WorkStatus;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import javax.jms.JMSException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+public class TibcoWorkFlowTaskTest {
+
+	private final static String caFile = "/cafile";
+
+	private final static String url = "ssl://localhost:7222";
+
+	private final static String username = "username";
+
+	private final static String passowrd = "password";
+
+	private final static String topic = "topic";
+
+	private final static String message = "test message";
+
+	private final static Tibjms tibjms = Mockito.mock(Tibjms.class);
+
+	private final static TibcoWorkFlowTask task = new TibcoWorkFlowTask(tibjms, url, caFile, username, passowrd);
+
+	@Before
+	public void setUp() {
+		task.setBeanName("Test");
+	}
+
+	@Test
+	public void executeMissingParameter() {
+		WorkContext ctx = getWorkContext(false);
+		WorkReport result = task.execute(ctx);
+		assertEquals(WorkStatus.FAILED, result.getStatus());
+		assertEquals(MissingParameterException.class, result.getError().getClass());
+	}
+
+	@Test
+	public void executeErrorInTibco() throws JMSException {
+		WorkContext ctx = getWorkContext(true);
+		doThrow(JMSException.class).when(tibjms).sendMessage(any(), any(), any(), any(), any(), any());
+		WorkReport result = task.execute(ctx);
+		assertEquals(WorkStatus.FAILED, result.getStatus());
+		assertEquals(JMSException.class, result.getError().getClass());
+	}
+
+	@Test
+	public void executeSuccess() throws JMSException {
+		WorkContext ctx = getWorkContext(true);
+		WorkReport result = task.execute(ctx);
+		assertEquals(WorkStatus.COMPLETED, result.getStatus());
+		verify(tibjms, times(1)).sendMessage(url, caFile, username, passowrd, topic, message);
+	}
+
+	private WorkContext getWorkContext(boolean withParams) {
+		WorkContext ctx = new WorkContext();
+		HashMap<String, String> map = new HashMap<>();
+		if (withParams) {
+			map.put("topic", topic);
+			map.put("message", message);
+		}
+
+		WorkContextDelegate.write(ctx, WorkContextDelegate.ProcessType.WORKFLOW_TASK_EXECUTION, task.getName(),
+				WorkContextDelegate.Resource.ARGUMENTS, map);
+		return ctx;
+	}
+
+}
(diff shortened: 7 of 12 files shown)


workflow-examples/README.md
```diff
@@ -46,6 +50,11 @@ Once the Jar has been compiled, ensure it's added to the pom.xml of the workflow
 	 <artifactId>workflow-examples</artifactId>
 	 <version>${parodos.version}</version>
 </dependency>
+<dependency>
+    <groupId>com.redhat.parodos</groupId>
```
masayag, line 54: shouldn't this be updated to `dev.parodos` ?

tasks/src/main/java/com/redhat/parodos/tasks/TibcoWorkFlowTask.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+
+@Slf4j
+public class TibcoWorkFlowTask extends BaseWorkFlowTask {
+
+	private Tibjms service;
+	private String url;
+	private String caFile;
+	private String username;
+	private String password;
```
gciavarrini, line 27: I'd suggest to make these `private final`

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.mockito.Mockito.*;
+
+public class TibcoWorkFlowTaskTest {
+
+	String caFile = "/cafile";
+	String url = "ssl://localhost:7222";
+	String username = "username";
+	String passowrd = "password";
+	String topic = "topic";
+	String message = "test message";
```
gciavarrini, line 28: I'd suggest to make this `private final`

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,73 @@
...
+	Tibjms tibjms = Mockito.mock(Tibjms.class);
+	TibcoWorkFlowTask task = new TibcoWorkFlowTask(tibjms, url, caFile, username, passowrd);
```
gciavarrini, line 30: It's a good practice to initialize objects in `@Before` method. 
In this way, the object is recreated before each test.

In addition, make these objects private as well.

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+	@Before
+	public void before() {
```
gciavarrini, line 33: To be consistent whit other tests, this method should be renamed in `setUp()`

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,73 @@
...
+		task.setBeanName("Test");
+	}
+	@Test
+	public void executeMissingParameter() {
+		WorkContext ctx = getWorkContext(false);
+		WorkReport result = task.execute(ctx);
+		assertEquals(WorkStatus.FAILED, result.getStatus());
+		assertEquals(MissingParameterException.class, result.getError().getClass());
```
gciavarrini, line 41: I think that this can be simplified as: 
`assertThrows(MissingParameterException.class, () -> result.getError());`

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	@Test
+	public void executeErrorInTibco() throws JMSException {
+		WorkContext ctx = getWorkContext(true);
+		doThrow(JMSException.class).when(tibjms).sendMessage(any(),any(),any(),any(),any(),any());
+		WorkReport result = task.execute(ctx);
+		assertEquals(WorkStatus.FAILED, result.getStatus());
+		assertEquals(JMSException.class, result.getError().getClass());
```
gciavarrini, line 50: same as before

workflow-examples/src/main/java/com/redhat/parodos/examples/TibcoWorkFlowConfiguration.java
```diff
@@ -0,0 +1,34 @@
+// @formatter:off
```
gciavarrini, line 1: why is it necessary to disable the formatter for the whole file?

gciavarrini (changes requested):

pom.xml
```diff
@@ -62,6 +62,7 @@
         <module>notification-service</module>
         <module>pattern-detection-library</module>
         <module>coverage</module>
+        <module>tasks</module>
```
nirarg, line 65: This module is going to be used for all common tasks
I think it should be renamed to more informative name
I suggest: `workflow-common-tasks`

tasks/src/main/java/com/redhat/parodos/tasks/Tibjms.java
nirarg, line 1: I suggest to keep this module directory only with the tasks files
Remove all other utils files to another sub-directory

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: why?

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: good practice? I don't think so. Not in this case. In the case of simple initialization, we have the Java language that works for us. You don't need to use additional libraries.
Why make them private? No one is going to use this class outside the test. The default is package-private.

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: it can not. I am not testing a method that throws.

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: see previous reply.

workflow-examples/src/main/java/com/redhat/parodos/examples/TibcoWorkFlowConfiguration.java
ydayagi: so I don't need to execute it for every change. I don't need this formatter.

pom.xml
ydayagi: We know it is common. We know it is for workflow. Why add all this text? I can somehow understand the addition of 'workflow-'. Some how ...

tasks/src/main/java/com/redhat/parodos/tasks/Tibjms.java
ydayagi: it only makes it more complex. I will move the whole tibco task to a sub-directory

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
gciavarrini: `private` = it's a best practice to keep the scope as narrow as you can.
`final` =  It's a good practice to declare final a variable to be sure that it won't be modified unintentionally, and in addition it makes the understanding of code easier.

TBH this strings can be declared `private static final` so we're stating that such variables are constants and that they are associated with the class (not the instance of the class)

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
gciavarrini: If you modify the object inside one of tests, then following tests will use a "dirty" object.
Regarding the `private` = It's a good practice to start to use the narrow privilege and the increase it if needed (so, `private` is more restrictive then `package-private`). Ideally, a good practice is to create immutable classes, but I understand that we can be flexible in this.

pom.xml
nirarg: We know its `common`, I'm not sure how clear it would be for new users, I think its better to make it explicit so it would more clear.
Anyway, I don't have any problem with changing to `workflow-tasks`

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: The Before does not solve it. It only happens once in the beginning. If test A changes something then test B will see a dirty obj.

workflow-examples/README.md
ydayagi: @masayag fixed

tasks/src/main/java/com/redhat/parodos/tasks/Tibjms.java
ydayagi: more hierarchy more problems more difficulties reading and following files

external-dependencies/download-tibco.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+if [ -f "tibjms.jar" ]; then exit 0; fi
+rm -rf /tmp/tibco
```
masayag, line 3: nit: maybe `mktemp -d tibco-XXX` ? 
(just in case there are few build processes running on the same machine)

pom.xml
```diff
@@ -62,6 +62,8 @@
         <module>notification-service</module>
         <module>pattern-detection-library</module>
         <module>coverage</module>
+        <module>prebuilt-tasks</module>
+        <module>external-dependencies</module>
```
masayag, line 66: nit: just for semantic, pls move this above `prebuilt-tasks`

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
```diff
@@ -0,0 +1,24 @@
+package com.redhat.parodos.tasks.tibco;
+
+import com.tibco.tibjms.TibjmsConnectionFactory;
+
+import javax.jms.*;
```
masayag, line 5: according to coding convention, `*` should be expanded

prebuilt-tasks/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+            <artifactId>external-dependencies</artifactId>
+            <version>${revision}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.24</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
```
masayag, line 31: I haven't seen a direct use of this dependency.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/Tibjms.java
```diff
@@ -0,0 +1,8 @@
+package com.redhat.parodos.tasks.tibco;
+
+public interface Tibjms {
```
masayag, line 3: nit: TibJMS ?

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
```diff
@@ -0,0 +1,24 @@
...
+		TibjmsConnectionFactory factory = new TibjmsConnectionFactory(url);
+		if (!caFile.isEmpty()) {
+			factory.setSSLTrustedCertificate(caFile);
+		}
+		Connection connection = factory.createConnection(username, password);
+		Session session = connection.createSession(javax.jms.Session.AUTO_ACKNOWLEDGE);
+		Destination destination = session.createTopic(topic);
+		MessageProducer producer = session.createProducer(destination);
+		TextMessage textMessage = session.createTextMessage(message);
+		producer.send(textMessage);
```
masayag, line 21: The created resources (e.g. connection) are auto-closable. However, What makes them close?

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,75 @@
...
+	public void executeSuccess() throws JMSException {
+		WorkContext ctx = getWorkContext(true);
+		WorkReport result = task.execute(ctx);
+		assertEquals(WorkStatus.COMPLETED, result.getStatus());
+		verify(tibjms, times(1)).sendMessage(url, caFile, username, passowrd, topic, message);
+	}
+
+	private WorkContext getWorkContext(boolean withParams) {
+		WorkContext ctx = new WorkContext();
+		Map map;
```
masayag, line 65: no need for type for the Map?

workflow-examples/README.md
```diff
@@ -42,10 +46,15 @@ Once the Jar has been compiled, ensure it's added to the pom.xml of the workflow
...
 
 <dependency>
-	 <groupId>com.redhat.parodos</groupId>
+	 <groupId>dev.parodos</groupId>
 	 <artifactId>workflow-examples</artifactId>
 	 <version>${parodos.version}</version>
 </dependency>
+<dependency>
+    <groupId>dev.parodos</groupId>
+    <artifactId>tasks</artifactId>
```
masayag, line 55: pls change to: prebuilt-tasks

workflow-examples/pom.xml
```diff
@@ -81,6 +81,12 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>dev.parodos</groupId>
+            <artifactId>prebuilt-tasks</artifactId>
+            <version>${revision}</version>
+            <scope>compile</scope>
```
masayag, line 88: `compile` is the default scope. It can be removed.

external-dependencies/download-tibco.sh
ydayagi: really? isn't each build in its own container?

masayag: I don't know. maybe @eloycoto knows?

pom.xml
ydayagi: definitely 'nit'

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
ydayagi: what do u mean?

ydayagi: I need many classes from that package. So I use '*'. That is super common.

prebuilt-tasks/pom.xml
ydayagi: I did not add it. I think it is for the '@slf4j'

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/Tibjms.java
ydayagi: comeon...tib is also an abbreviation. This is how they do it. LEt's keep it

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
ydayagi: if it is auto-closable then what is the problem? I took this from TIBCO example/docs

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
ydayagi: no

tasks/src/main/java/com/redhat/parodos/tasks/TibcoWorkFlowTask.java
ydayagi: @gciavarrini fixed

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: @gciavarrini fixed

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: fixed

tasks/src/test/java/com/redhat/parodos/tasks/TibcoWorkFlowTaskTest.java
ydayagi: @gciavarrini fixed

workflow-examples/src/main/java/com/redhat/parodos/examples/TibcoWorkFlowConfiguration.java
ydayagi: @gciavarrini fixed

pom.xml
ydayagi: @nirarg fixed

workflow-examples/README.md
ydayagi: @masayag fixed

prebuilt-tasks/pom.xml
masayag: '@slf4j' comes from lombok. pls try to build without that dependency and see if it is actually being used.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTask.java
```diff
@@ -0,0 +1,76 @@
...
+		return List.of(WorkFlowTaskOutput.OTHER);
+	}
+
+	@Override
+	public WorkReport execute(WorkContext workContext) {
+		try {
+			String topic = getParameterValue(workContext, "topic");
+			String message = getParameterValue(workContext, "message");
+
+			log.debug("sending to topic " + topic + " message: " + message);
```
masayag, line 63: instead of string concatenation, use 
```
log.debug("sending to topic {} message: {}", topic, message);
```

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
masayag: there are coding conventions to this project. we should respect them, else, each one will use its own style.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
masayag: well, auto-closable works in try-with-resource which isn't the case here.
I'm afraid it may end up in leaking resources.

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
gciavarrini: +1 Moti
In truth, this generates the warning "Raw use of parameterized class".
You must provide the proper generic value inside diamonds.

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,75 @@
...
+		if ( withParams == true ) {
```
gciavarrini, line 66: This can be simplified `if(withParams)`

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,75 @@
...
+			map = Map.of("topic",topic,"message",message);
+		} else {
+			map = new HashMap();
+		}
```
gciavarrini, line 70: Why do not create the HashMap before the `if` and then add value if `withParams` is true ?
```java
HashMap<String, String> map = new HashMap<>();
if(withParams){
    map.put("topic", topic);
    map.put("message", message);
}
```

In addition, `map` is a variable name too generic, what about `messageMap` or something more "declarative"?

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTask.java
```diff
@@ -0,0 +1,76 @@
...
+		this.service = service;
+		this.url = url;
+		this.caFile = caFile;
+		this.username = username;
+		this.password = password;
+	}
+
+	@Override
+	public @NonNull List<WorkFlowTaskParameter> getWorkFlowTaskParameters() {
+		List params = new LinkedList<WorkFlowTaskParameter>();
```
gciavarrini, line 44: "Raw use of parameterized class" warning here. Please replace this with `LinkedList<WorkFlowTaskParameter> params = new LinkedList<>();`

workflow-examples/src/main/java/com/redhat/parodos/examples/TibcoWorkFlowConfiguration.java
gciavarrini: It is still there. Can you double check?

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,75 @@
+// @formatter:off
+package com.redhat.parodos.tasks.tibco;
+
+import com.redhat.parodos.tasks.tibco.TibcoWorkFlowTask;
+import com.redhat.parodos.tasks.tibco.Tibjms;
```
gciavarrini, line 5: unused imports, please remove them.

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.mockito.Mockito.*;
+
+public class TibcoWorkFlowTaskTest {
+
+	private final String caFile = "/cafile";
+	private final String url = "ssl://localhost:7222";
+	private final String username = "username";
+	private final String passowrd = "password";
+	private final String topic = "topic";
+	private final String message = "test message";
```
gciavarrini, line 30: missing `static` as per our previous discussion :)

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibjmsImpl.java
ydayagi: conventions where? look at Java code all over the world. People prefer '*' then specifying a gazillion import lines. Besides, who says others are right. I will change it just for passing the PR. But this is no doubt the wrong thing to do.

ydayagi: by the way, when you tell intellij to import a class then this is what intellij does. Do we know better? we know better than them what is the convention?

prebuilt-tasks/pom.xml
ydayagi: fixed

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
ydayagi: this is standard java usage. It makes code simpler and less type independent where possible. So when you change the map type you want to use you change only where it is created and only where necessary. That's the whole idea of interfaces and generics (templates). But again...just for merging the PR I will change it. BTW...intellij does that automatically. Not my idea.

ydayagi: I don't see a warning

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
ydayagi: fixed

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
ydayagi: why? see my previous comment. I changed it just so we can continue

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTaskTest.java
ydayagi: fixed

masayag:
lgtm

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/tibco/TibcoWorkFlowTask.java
ydayagi: and that's why?

ydayagi: that's what preventing approval of this PR?

pkliczewski:
@gciavarrini please review

gciavarrini (approved):
/lgtm

PR #129, opened 2023-03-20
Decouple examples from service

masayag:
The purpose of this PR is to decouple the user's workflows from being built into the workflow-service by using the PropertiesLauncher as the Start-Class of the application.

Extra dependencies can be provided via env-var `loader.path`. In addition, to satisfy all runtime dependencies, the workflow-example is being built with a plugin to include all of its dependencies in a single jar.

For local development, a local profile was added to include the
workflow-examples dependency in the service's path:

```
mvn clean install -Plocal
```

### DIFF
--- a/workflow-service/pom.xml
+++ b/workflow-service/pom.xml
@@ -126,11 +126,6 @@
             <artifactId>workflow-engine</artifactId>
             <version>${revision}</version>
         </dependency>
-        <dependency>
-            <groupId>dev.parodos</groupId>
-            <artifactId>workflow-examples</artifactId>
-            <version>${revision}</version>
-        </dependency>
         <!-- END Parodos Library Dependencies -->
         <!-- START Developer Productivity Dependencies -->
         <dependency>
@@ -214,6 +209,9 @@
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
                 <version>${springframework.boot.version}</version>
+                <configuration>
+                    <layout>ZIP</layout>
+                </configuration>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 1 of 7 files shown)


workflow-service/pom.xml
```diff
@@ -126,11 +126,6 @@
             <artifactId>workflow-engine</artifactId>
             <version>${revision}</version>
         </dependency>
-        <dependency>
```
RichardW98, line 129: is there any way to keep it for local profile, and remove it for other profiles, e.g. dev & release ?  so it will be easier for our local debug

gciavarrini (commented):
/lgtm

workflow-service/pom.xml
pkliczewski: Yes, we can add new profile [here](https://github.com/redhat-developer/parodos/blob/main/pom.xml#L189)

RichardW98: Great! thank you Piotr

masayag: added

RichardW98 (approved):

pkliczewski:
@masayag we fail with building the image. Please take a look

> Service 'workflow-service' failed to build : Build failed

pkliczewski (commented):
/lgtm

workflow-service/start_workflow_service.sh
```diff
@@ -1 +1,3 @@
-java -jar -Dspring.profiles.active=local target/workflow-service-1.0.3-SNAPSHOT.jar
+java -jar -Dspring.profiles.active=local \
```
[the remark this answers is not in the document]
masayag: Do you suggest deleting this script? It also repeats itself under workflow-examples.
And if we maintain that script, it should probably be parameterized with the version instead hard-coding it

PR #139, opened 2023-03-22
FLPATH-171: Generate SDK for workflow-service

gciavarrini:
Relates to PR #119 

Generate SDK in the `workflow-service-sdk` module from workflow-service openAPI JSON file.

First, enable the generation of openAPI JSON file in the `workflow-service` module.
Then, generate SDK from such JSON file in the `workflow-service-sdk` module.
Finally, an example of usage in the `workflow-example` package has been provided.


gciavarrini:
Open questions from PR #119 discussion:
* Should the generated code be added to this PR? (@masayag @pkliczewski )

masayag:
> Open questions from PR #119 discussion:
> 
> * Should the generated code be added to this PR? (@masayag @pkliczewski )

The output of the SDK will be a jar containing the generated classes.
I think we should include those classes in the code base (they will be overridden each time the API changes).
Without having the classes, the only way to debug is by decompile the SDK's classes.
We should exclude the generated classes from the formatter/checkstyle/code coverage and other code analysis.

gciavarrini:
> > Open questions from PR #119 discussion:
> > 
> > * Should the generated code be added to this PR? (@masayag @pkliczewski )
> 
> The output of the SDK will be a jar containing the generated classes. I think we should include those classes in the code base (they will be overridden each time the API changes). Without having the classes, the only way to debug is by decompile the SDK's classes. We should exclude the generated classes from the formatter/checkstyle/code coverage and other code analysis.

Sounds good to me. I'll add them.
I've already disabled the `pring-javaformat-maven-plugin` ([here](https://github.com/redhat-developer/parodos/blob/018c3be5e7f52a9b80b598f6ef62838ea12c0804/workflow-service-sdk/pom.xml#L31)) because it made the build fail.

gciavarrini:
> > Sounds good to me. I'll add them. I've already disabled the `pring-javaformat-maven-plugin` ([here](https://github.com/redhat-developer/parodos/blob/018c3be5e7f52a9b80b598f6ef62838ea12c0804/workflow-service-sdk/pom.xml#L31)) because it made the build fail.
> 
> We might also need to update the code-coverage something to ignore those classes. At some point something will complain about it :-)

Added generated code in this commit https://github.com/redhat-developer/parodos/pull/139/commits/82378ce88cdb413b4b576bcdbdac7d21e29ad6d4 

Added `exclude` configuration in `coverage` [pom](https://github.com/redhat-developer/parodos/blob/ee4e78c0e5281e02d6dd7d62cda503128da1d81e/coverage/pom.xml#L61-L63)

pkliczewski:
@gciavarrini please rebase

openapi/api-docs.json
nirarg, line 1: I think this file should be located under `workflow-service/openapi` (what do you say even about adding `generated` to the path? to make it clear that it auto generated file - `workflow-service/generated/openapi`)

workflow-service/pom.xml
```diff
@@ -210,18 +205,102 @@
...
+                <groupId>org.springdoc</groupId>
+                <artifactId>springdoc-openapi-maven-plugin</artifactId>
+                <version>1.1</version>
+                <configuration>
+                    <outputFileName>api-docs.json</outputFileName>
+                    <outputDir>${project.parent.basedir}/openapi</outputDir>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>integration-test</id>
```
nirarg, line 283:
```suggestion
                        <id>generate-openapi-json</id>
```

gciavarrini: addressed

openapi/api-docs.json
gciavarrini: addressed

pkliczewski:
@gciavarrini please rebase

PR #142, opened 2023-03-22
Workflow-service: Fix migrations issues on integration test

eloycoto:


pkliczewski (commented):
/lgtm

RichardW98 (approved):

workflow-service/src/main/resources/db/changelog/data.xml
```diff
@@ -4,7 +4,7 @@
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
 
-    <changeSet id="1" author="luke" runAlways="true" runOnChange="true">
+    <changeSet id="1" author="luke" runAlways="false" runOnChange="true">
```
gciavarrini, line 7: Can you explain me a bit what this flag does? Why this flag fix the issue?

gciavarrini: Oh ok! Thanks for the clarification!

eloycoto: Ok, as I said, the issue is that it is trying to create the table users again each time that the service starts:

```
2023-03-21T15:10:59.990365102Z stdout F 
2023-03-21T15:10:59.990369202Z stdout F org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set db/changelog/tables.xml::1::luke:
2023-03-21T15:10:59.990373302Z stdout F      Reason: liquibase.exception.DatabaseException: ERROR: relation "user" already exists [Failed SQL: (0) CREATE TABLE public."user" (id UUID DEFAULT gen_random_uuid() NOT NULL, username VARCHAR(255), password VARCHAR(255), authority VARCHAR(255), enabled SMALLINT, first_name VARCHAR(255), last_name VARCHAR(255), email VARCHAR(255), create_date TIMESTAMP WITHOUT TIME ZONE DEFAULT NOW(), modify_date TIMESTAMP WITHOUT TIME ZONE DEFAULT NOW(), object_version BIGINT, CONSTRAINT USER_PKEY PRIMARY KEY (id))]
```

Consecuentily, if the pod is restarted for any reason (Healthcheck, or process died) the second time, it cannot start again due to trying to create tables again in the database, and those are already created. 


The reason to increase the ReadinessProbe is to give it more time to start and not restart workflow-service pod in the middle of initialization, so the issue is not fixed, it's alleviated due to fewer restarts and that we bring more time to start the service.  

With this change, the issue of creating tables is now fixed, *But* sadly, another issue is hit, so when the workflow is registered, hits another DB constraint, something that I'm investigating.

PR #143, opened 2023-03-22
Update workflow-service README

masayag:


workflow-service/README.md
```diff
@@ -22,10 +22,19 @@ test the service. It should be removed once actual Tasks and WorkFlows start get
 To start the application run the following from the root folder of 'workflow-service'.
 
 ```shell
-
-java -Dspring.profiles.active=local -jar target/workflow-service-0.0.1-SNAPSHOT.jar
-
+VERSION=1.0.3-SNAPSHOT
```
pkliczewski, line 25: Do you want use to change this version every time we release?

lshannon: Yes, we need too

masayag: > Do you want use to change this version every time we release?

TBH, no - I'd rather have a template of this script and generate it based on the version each time. Let maintenance burden on us.

PR #146, opened 2023-03-22
FLPATH-204: cleanup definitions in db before loading new beans

RichardW98:
[FLPATH-204](https://issues.redhat.com/browse/FLPATH-204)

update database with new workflow configuration after app restarts

*note:   executions and project table are not cleaned


pkliczewski (commented):
looks good, please add a test

RichardW98:
tests added

masayag:
lgtm

RichardW98:
> What will happen if there are tasks execution or workflow execution in the DB ? Wouldn't it cause reference key constrain violation? Perhaps there is a need to check there are not executions of any kind before deleting?

yes, the executions will be deleted as well if its definition is removed. I will change the logic from removing definitions to updating/adding

PR #149, opened 2023-03-23
OWNERS: Add more reviewers

nirarg:


nirarg:
Added @rgolangh as collaborator (READ access)
If the READ access is enough, I suggest to change most collaborators to this access level

PR #153, opened 2023-03-23
Enable graceful shutdown endpoint

masayag:
The PR enables the shutdown endpoint which can be triggered by:
(for `local` profile)
```
curl -u test:test -X POST  localhost:8080/actuator/shutdown
```

In addition, a script for local invocation of the `workflow-service` is added. If the service dies, the script will relaunch it. It may be useful for testing/development locally when there is a need to replace only the workflows jar.

The graceful shutdown timeout (`spring.lifecycle.timeout-per-shutdown-phase`) is set to 25s, since the default timeout for K8s pods is 30s. This can be revisited and tuned by changing the `workflow-service` deployment to specify a different timeout by terminationGracePeriodSeconds.

The graceful shutdown sequence should look as:
```
2023-03-27 22:54:55.497  INFO 861119 --- [       Thread-7] o.s.b.w.e.tomcat.GracefulShutdown        : Commencing graceful shutdown. Waiting for active requests to complete
2023-03-27 22:54:55.507  INFO 861119 --- [tomcat-shutdown] o.s.b.w.e.tomcat.GracefulShutdown        : Graceful shutdown complete
2023-03-27 22:54:55.571  INFO 861119 --- [       Thread-7] c.r.p.w.e.service.WorkFlowServiceImpl    :  >>>>>>>>> Shutting down the workflow service <<<<<<<<<<<
2023-03-27 22:54:55.571  INFO 861119 --- [       Thread-7] c.r.p.w.r.BeanWorkFlowRegistryImpl       :  >>>>>>>>> Shutting down the bean registry <<<<<<<<<<<
2023-03-27 22:54:55.572  INFO 861119 --- [       Thread-7] j.LocalContainerEntityManagerFactoryBean : Closing JPA EntityManagerFactory for persistence unit 'default'
2023-03-27 22:54:55.574  INFO 861119 --- [       Thread-7] com.zaxxer.hikari.HikariDataSource       : ${DATASOURCE_NAME} - Shutdown initiated...
2023-03-27 22:54:55.578  INFO 861119 --- [       Thread-7] com.zaxxer.hikari.HikariDataSource       : ${DATASOURCE_NAME} - Shutdown completed.
```

### DIFF
--- a/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
@@ -43,6 +43,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 import org.springframework.web.server.ResponseStatusException;
 
+import javax.annotation.PreDestroy;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
@@ -242,6 +243,11 @@ public class WorkFlowServiceImpl implements WorkFlowService {
 
 		// TODO: validate required parameters from definition
 		return null;
+	}
+
+	@PreDestroy
+	public void gracefulShutdown() {
+		log.info(">> Shutting down the workflow service");
 	}
 
 }
--- a/workflow-examples/start_workflow_service.sh
+++ b/workflow-examples/start_workflow_service.sh
@@ -1,5 +1,4 @@
-export SERVER_PORT=8080
+pushd ../workflow-service
+./start_workflow_service.sh
+popd
 
-java -jar -Dspring.profiles.active=local \
-    -Dloader.path=../workflow-examples/target/workflow-examples-1.0.5-SNAPSHOT-jar-with-dependencies.jar \
-    ../workflow-service/target/workflow-service-1.0.5-SNAPSHOT.jar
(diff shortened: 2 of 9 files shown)


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
```diff
@@ -359,4 +360,9 @@ private String validateWorkflow(String workflowName, WorkFlow workFlow) {
 		return null;
 	}
 
+	@PreDestroy
+	public void gracefulShutdown() {
+		log.info(" >>>>>>>>> Shutting down the workflow service <<<<<<<<<<<");
```
masayag, line 365: @lshannon @RichardW98 @anludke which logic would we like to have in this context?
Right before the application ends, there are no active requests, but there may be running executions.

RichardW98: yes we need to wait for running tasks and their persistence to complete before shutdown, and prevent upcoming workflow/task runs

masayag:
> Do you think it makes sense to add shutdown endpoint to notification service?

Could be. Or at least enable a graceful shutdown to it.
I suggest tracking it separately since it wasn't evaluated as part of the workflow-service.

PR #155, opened 2023-03-23
Flpath 204: update database with new workflow configuration after app restarts

RichardW98:
[FLPATH-204](https://issues.redhat.com/browse/FLPATH-204)

all related executions will be removed if their workflow definition is cleaned up. this pr is to update workflowdefinition if it exists and add new ones if not. No deletion will happen

add missing author name to source code files

### DIFF
--- a/workflow-examples/src/main/java/org/parodos/workflow/examples/task/SimpleWorkFlowCheckerTask.java
+++ b/workflow-examples/src/main/java/org/parodos/workflow/examples/task/SimpleWorkFlowCheckerTask.java
@@ -1,0 +1,34 @@
+package org.parodos.workflow.examples.task;
+
+import com.redhat.parodos.workflow.task.checker.BaseWorkFlowCheckerTask;
+import com.redhat.parodos.workflow.task.enums.WorkFlowTaskOutput;
+import com.redhat.parodos.workflow.task.parameter.WorkFlowTaskParameter;
+import com.redhat.parodos.workflows.work.DefaultWorkReport;
+import com.redhat.parodos.workflows.work.WorkContext;
+import com.redhat.parodos.workflows.work.WorkReport;
+import com.redhat.parodos.workflows.work.WorkStatus;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Collections;
+import java.util.List;
+
+@Slf4j
+public class SimpleWorkFlowCheckerTask extends BaseWorkFlowCheckerTask {
+
+	@Override
+	public WorkReport checkWorkFlowStatus(WorkContext workContext) {
+		log.info("NamespaceApprovalWorkFlowCheckerTask");
+		return new DefaultWorkReport(WorkStatus.FAILED, workContext);
+	}
+
+	@Override
+	public List<WorkFlowTaskParameter> getWorkFlowTaskParameters() {
+		return Collections.emptyList();
+	}
+
+	@Override
+	public List<WorkFlowTaskOutput> getWorkFlowTaskOutputs() {
+		return Collections.emptyList();
+	}
+
+}
--- a/workflow-examples/src/main/java/org/parodos/workflow/examples/task/SimpleWorkFlowCheckerTask.java
+++ b/workflow-examples/src/main/java/org/parodos/workflow/examples/task/SimpleWorkFlowCheckerTask.java
@@ -1,34 +1,19 @@
 package org.parodos.workflow.examples.task;
 
 import com.redhat.parodos.workflow.task.checker.BaseWorkFlowCheckerTask;
-import com.redhat.parodos.workflow.task.enums.WorkFlowTaskOutput;
-import com.redhat.parodos.workflow.task.parameter.WorkFlowTaskParameter;
 import com.redhat.parodos.workflows.work.DefaultWorkReport;
 import com.redhat.parodos.workflows.work.WorkContext;
 import com.redhat.parodos.workflows.work.WorkReport;
 import com.redhat.parodos.workflows.work.WorkStatus;
 import lombok.extern.slf4j.Slf4j;
 
-import java.util.Collections;
-import java.util.List;
-
 @Slf4j
 public class SimpleWorkFlowCheckerTask extends BaseWorkFlowCheckerTask {
 
 	@Override
 	public WorkReport checkWorkFlowStatus(WorkContext workContext) {
-		log.info("NamespaceApprovalWorkFlowCheckerTask");
+		log.info("SimpleWorkFlowCheckerTask");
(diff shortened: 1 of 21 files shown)


workflow-service/src/main/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceImpl.java
```diff
@@ -98,38 +99,73 @@ public WorkFlowDefinitionResponseDTO save(String workFlowName, WorkFlowType work
...
 
+		String stringifyParameters = WorkFlowDTOUtil
+				.writeObjectValueAsString(convertWorkFlowParameters(workFlowParameters));
 		// set and save workflow definition
-		WorkFlowDefinition workFlowDefinition = workFlowDefinitionRepository.save(WorkFlowDefinition.builder()
-				.name(workFlowName).type(workFlowType).createDate(new Date())
-				.parameters(WorkFlowDTOUtil.writeObjectValueAsString(convertWorkFlowParameters(workFlowParameters)))
-				.modifyDate(new Date()).numberOfWorks(works.size()).processingType(workFlowProcessingType.name())
-				.build());
+		WorkFlowDefinition workFlowDefinition = workFlowDefinitionRepository.findFirstByName(workFlowName);
```
gciavarrini, line 105: May I ask where the function `findFirstByName` has been implemented?

workflow-service/src/test/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceDataTest.java
```diff
@@ -69,10 +90,10 @@ void cleanAllDefinition_should_returnEmpty_when_DatabaseIsCleaned() {
 		assertThat(workFlowCheckerMappingDefinitionRepository.findAll()).hasSize(1);
 		assertThat(workFlowWorkRepository.findAll()).hasSize(1);
 
-		workFlowDefinitionService.cleanAllDefinitions();
+		workFlowDefinitionService.cleanAllDefinitionMappings();
```
gciavarrini, line 93: The method name is `cleanAllDefinition_should_returnEmpty_when_DatabaseIsCleaned` can you update it?
I mean before this change, all the methods were supposed to be empty now only `workFlowCheckerMappingDefinitionRepository` and `workFlowWorkRepository` are empty.

workflow-service/src/test/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceImplTest.java
```diff
@@ -109,6 +131,42 @@ public void simpleSaveTest() {
...
+				WorkFlowType.ASSESSMENT, Collections.emptyList(), List.of(workFlowTask),
+				WorkFlowProcessingType.SEQUENTIAL);
+
+		// then
+		assertNotNull(workFlowDefinitionResponseDTO);
+		assertNotNull(workFlowDefinitionResponseDTO.getId());
+		assertEquals(workFlowDefinitionResponseDTO.getName(), workFlowName);
+
+		Mockito.verify(this.workFlowDefinitionRepository, Mockito.times(1)).save(any());
+		Mockito.verify(this.workFlowTaskDefinitionRepository, Mockito.times(0)).save(any());
```
gciavarrini, line 167: nit: I'd suggest to user `Mockito.never()` to improve the test readability

workflow-service/src/test/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceImplTest.java
```diff
@@ -109,6 +131,42 @@ public void simpleSaveTest() {
...
+		WorkFlowDefinitionResponseDTO workFlowDefinitionResponseDTO = this.workFlowDefinitionService.save(workFlowName,
+				WorkFlowType.ASSESSMENT, Collections.emptyList(), List.of(workFlowTask),
+				WorkFlowProcessingType.SEQUENTIAL);
+
+		// then
+		assertNotNull(workFlowDefinitionResponseDTO);
+		assertNotNull(workFlowDefinitionResponseDTO.getId());
+		assertEquals(workFlowDefinitionResponseDTO.getName(), workFlowName);
+
+		Mockito.verify(this.workFlowDefinitionRepository, Mockito.times(1)).save(any());
```
gciavarrini, line 166: Do you think It's possible to assert the parameter of `save()` as well? I mean to use `eq(..)` instead of `any()` ? In this way, we can test that there aren't unwanted edit to the object.

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceImpl.java
```diff
@@ -194,6 +230,8 @@ public void saveWorkFlowChecker(String workFlowTaskName, String workFlowCheckerN
 							.findFirstByCheckWorkFlow(checkerWorkFlowDefinitionEntity))
 					.orElse(WorkFlowCheckerMappingDefinition.builder().checkWorkFlow(checkerWorkFlowDefinitionEntity)
 							.cronExpression(workFlowCheckerDTO.getCronExpression()).tasks(new ArrayList<>()).build());
+			WorkFlowTaskDefinition workFlowTaskDefinition = workFlowTaskDefinitionRepository
```
gciavarrini, line 233: Can you double check this? `workFlowTaskDefinition` is never used: can be this a bug?

workflow-examples/src/main/java/org/parodos/workflow/examples/task/SimpleWorkFlowCheckerTask.java
```diff
@@ -0,0 +1,34 @@
...
+
+	@Override
+	public List<WorkFlowTaskParameter> getWorkFlowTaskParameters() {
+		return Collections.emptyList();
+	}
+
+	@Override
+	public List<WorkFlowTaskOutput> getWorkFlowTaskOutputs() {
+		return Collections.emptyList();
+	}
```
gciavarrini, line 32: nit: no need to override these methods because they are already implemented here https://github.com/redhat-developer/parodos/blob/de2451a3e499bf2ea31e5eeae221e3c5782c00b4/parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/WorkFlowTask.java#L44-L57

gciavarrini (commented):
This PR lgtm.

For the moment, I don't add the label to have another review to be sure that everything is ok

PR #163, opened 2023-03-24
Change OpenAPI filename to `openapi.json`

gciavarrini:
According to OpenAPI specifications filename should be `openapi.json`
    

> Document Structure
> An OpenAPI document MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the author. In the latter case, Reference Objects and Schema Object $ref keywords are used.
> It is RECOMMENDED that the root OpenAPI document be named: openapi.json or openapi.yaml.

    
See https://swagger.io/specification/


eloycoto:
Quick thing, the workflow-sdk is not affected by this change?

gciavarrini:
> Quick thing, the workflow-sdk is not affected by this change?

yes, I updated the sdk-pom as well [here](https://github.com/redhat-developer/parodos/blob/70952f57d5592b34375924110edbd96eb19dbece/workflow-service-sdk/pom.xml#L80)

eloycoto:
image
Have a good weekend  and thanks :-)

PR #175, opened 2023-03-28
[WIP] - Enable graceful shutdown sequence

masayag:


masayag:
@RichardW98 I created this PR to continue the discussion that was started [here](https://github.com/parodos-dev/parodos/pull/153#discussion_r1146506187).

How would you suggest preventing upcoming tasks?
For workflow, with or without this PR, any requests to tomact will be denied.
However, for tasks, it means there is a thread already running in the system and executing the workflow.
I'm not sure how to identify point-of-time in which we can prevent moving to the next task in the workflow.
Do you have any hints on this?

PR #179, opened 2023-03-29
Fix: change workflow-service openapi generator phases

gciavarrini:
SpringBoot must be started/stopped to create the `openapi.json` file.
Current selected phases generated the file and stopped SpringBoot wrongly.


pkliczewski (commented):
/approve

PR #183, opened 2023-03-29
Version bump

pkliczewski:


notification-service/pom.xml
```diff
@@ -129,6 +138,53 @@
         	<groupId>org.springframework.boot</groupId>
         	<artifactId>spring-boot-configuration-processor</artifactId>
         	<optional>true</optional>
+        </dependency>
```
pkliczewski, line 141: @nirarg This should not be there but release profile fails to build java docs for sdk. Please investigate why it is needed here.

pkliczewski: Let's fix it in different PR

RichardW98 (approved):

PR #184, opened 2023-03-30
FLPATH-207: Add IN_PROGRESS Work Status

nirarg:
In case of Checker not completed return IN_PROGRESS status instead of FAILED


pkliczewski (commented):
/lgtm

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/aspect/WorkFlowExecutionAspect.java
```diff
@@ -258,7 +258,7 @@ private WorkReport handlePostWorkflowExecution(boolean isMaster, WorkFlowExecuti
 						|| !WorkFlowStatus.COMPLETED.equals(checkerExecution.getStatus()))) {
 			log.info("fail workflow: {} because it has pending/running checker(s)", workFlowDefinition.getName());
 			workFlowExecution.setStatus(WorkFlowStatus.PENDING);
-			report = new DefaultWorkReport(WorkStatus.FAILED, workContext);
+			report = new DefaultWorkReport(WorkStatus.IN_PROGRESS, workContext);
```
RichardW98, line 261: in_progress won't stop the workflow to wait for checker : https://github.com/parodos-dev/parodos/blob/main/workflow-engine/src/main/java/com/redhat/parodos/workflows/workflow/SequentialFlow.java#L62

RichardW98: but this pr is very meaningful as the  status enums in workflow-engine is apparently not enough.

RichardW98: @anludke @lshannon  please suggest if we can modify base workflow-engine library, and change its logic of "completed" for all types of workflow, e.g parallel, conditional, repeat, sequential ....

RichardW98 (approved):

PR #191, opened 2023-03-31
FLPATH-216: Add lombok annotations to service SDK generated files

gciavarrini:
Add `Lombok` annotations `@Data @AllArgsConstructor @Builder` to workflow-service SDK files.

The `UseSDKExample` is updated to demonstrate the use of the new available methods.

### DIFF
--- a/workflow-service-sdk/pom.xml
+++ b/workflow-service-sdk/pom.xml
@@ -19,6 +19,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <springframework.boot.version>2.5.14</springframework.boot.version>
         <openapi.generator.version>5.4.0</openapi.generator.version>
+        <lombok.version>1.18.26</lombok.version>
     </properties>
 
     <build>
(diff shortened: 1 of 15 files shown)


pkliczewski (commented):
/lgtm

workflow-service-sdk/pom.xml
```diff
@@ -173,7 +175,12 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>${lombok.version}</version>
+            <scope>provided</scope>
```
masayag, line 182: isn't this needed for runtime of the SDK?

gciavarrini: This is what Lombok [website](https://projectlombok.org/setup/maven) suggests to configure the dependency:

> To set up lombok with any build tool, you have to specify that the lombok dependency is required to compile your source code, but does not need to be present when running/testing/jarring/otherwise deploying your code. Generally this is called a 'provided' dependency.

Do you think that in our project the scope should be different?

masayag: No. Provided makes sense based on the description.

PR #194, opened 2023-04-02
Refactor workflow aspect

masayag:
WorkFlowExecutionAspect was refactored into several classes to allow separation and encapsulation of pre-execution and post-execution.

PR #197, opened 2023-04-03
FLPATH-41: Make continue service workflows execution asynchronous

masayag:
This PR changes the invocation of workflows that should be continued in async way to enable parallel execution of workflows in the system.

### DIFF
--- a/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/continuation/AsyncWorkFlowContinuer.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/continuation/AsyncWorkFlowContinuer.java
@@ -1,0 +1,13 @@
+package com.redhat.parodos.workflow.execution.continuation;
+
+import java.util.UUID;
+
+import com.redhat.parodos.workflows.work.WorkContext;
+import org.springframework.scheduling.annotation.Async;
+
+public interface AsyncWorkFlowContinuer {
+
+	@Async
+	void executeAsync(String projectId, String workflowName, WorkContext workContext, UUID executionId);
+
+}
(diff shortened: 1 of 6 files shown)


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/continuation/AsyncWorkFlowContinuer.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.parodos.workflow.execution.continuation;
+
+import java.util.UUID;
+
+import com.redhat.parodos.workflows.work.WorkContext;
+
+public interface AsyncWorkFlowContinuer {
+
+	void execute(String projectId, String workflowName, WorkContext workContext, UUID executionId);
```
rgolangh, line 9: The @Async annotation hides the fact that this method is called asynchronously. 
So my next thought is why not return a `Future<WorkReport>` in the signature to express that?

masayag: There are two implementations of the Async method. The other one expects to return a ComputableFuture if you need to return a handle for the result.
In this context, we don't need to return any response or a feedback. Just to spawn the execution and move to the next one.

RichardW98: what about move/copy the @Async from impl to here and rename the method to executeAsync(...)?

RichardW98:
can we make this one async too? https://github.com/RichardW98/parodos/blob/main/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java#L72

masayag:
> can we make this one async too? https://github.com/RichardW98/parodos/blob/main/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java#L72

For sure :-) 
For that I need https://github.com/parodos-dev/parodos/pull/194 to be merged first since the persistence of the workflow-execution to the DB is done at that point.
Then executing the workflow in its thread will have to be done from the workflow-aspect.

eloycoto (approved):

PR #199, opened 2023-04-03
Workflow: Properties annotation

eloycoto:
This commit introduces a new way to annotate workflow metadata to the workflow itself, the annotation comes with a single option, but with the idea to be expanded in the near future

Each workflow now has a way to define a version that is stored in the workflowDefinition.Properties column (JsonB) object. Users may want to determine the version by hand, or using any tool, like the example given using git-commit-id-plugin.

As an example:

```

@Bean(name = "simpleSequentialWorkFlow" + WorkFlowConstants.INFRASTRUCTURE_WORKFLOW)
@Infrastructure
@WorkFlowProperties(version = "${git.commit.id}")
WorkFlow simpleSequentialWorkFlowTask(@Qualifier("restCallTask") RestAPIWorkFlowTask restCallTask,
        @Qualifier("loggingTask") LoggingWorkFlowTask loggingTask) {
```

or:

```

@WorkFlowProperties(version = "v1.0.0")
```

The version is a string, but maybe we can think on semver here. I'm very open here, but the idea was to use the commit id AFAIK, and not sure if it adds value having in semver object.

The main reason to use Properties is that they can be extended with more fields like: `@WorkFlowProperties(version = "v1.0.0", metricName="myAwesomeWorkflow")` or some special logic like on_update, on_die, priority_key, labels.

Fix [FLPATH-17](https://issues.redhat.com//browse/FLPATH-17)


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/aspect/WorkFlowPropertiesAspect.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+import org.springframework.core.env.Environment;
+
+@Aspect
+@Component
+@Slf4j
+@Order(-1)
+public class WorkFlowPropertiesAspect {
```
eloycoto, line 21: There is no test for this yet, I need to figure it out some reflection issue

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/aspect/WorkFlowPropertiesAspect.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	public Object setProperitesForWorkflow(WorkFlow workFlow, WorkFlowProperties properties) throws IOException {
+		if (properties == null) {
+			return workFlow;
+		}
+		String version = env.resolvePlaceholders(properties.version());
+		if (version.isEmpty()) {
+			version = WorkFlowVersionServiceImpl.GetVersionHashForObject(workFlow);
+		}
```
gabriel-farache, line 50: Why do you want to force a version on the workflow? That will be a bit confusing as a user as the default value is empty string on the annotation, I would expect to have a, empty version in the DB.
Maybe the default version should be the git commit sha?

If you want to keep this behaviour with a generated version based on the hash of the object, it would be great to have it documented somewhere

parodos-model-api/src/main/java/com/redhat/parodos/workflow/annotation/WorkFlowProperties.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+
+@Target({ METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface WorkFlowProperties {
+
+	String version() default "";
```
gabriel-farache, line 15: Why not ${git.commit.id} by default?

gabriel-farache (commented):
Very nicely done, left a few comments.

And would be good to have some piece of documentation somewhere explaining where to use and how it's working (like you've done in the description of the PR)

workflow-examples/pom.xml
```diff
@@ -102,6 +107,32 @@
     </dependencies>
     <build>
         <plugins>
+            <plugin>
+                <groupId>pl.project13.maven</groupId>
+                <artifactId>git-commit-id-plugin</artifactId>
```
pkliczewski, line 112: Can we use tag if available?

workflow-engine/src/main/java/com/redhat/parodos/workflows/workflow/WorkFlowPropertiesMetadata.java
```diff
@@ -0,0 +1,18 @@
+package com.redhat.parodos.workflows.workflow;
+
+import lombok.Builder;
+
+@Builder
```
pkliczewski, line 5: What about generating getter and setter?

workflow-service/src/main/java/com/redhat/parodos/project/controller/ProjectController.java
```diff
@@ -82,6 +82,7 @@ public ResponseEntity<ProjectResponseDTO> createProject(@Valid @RequestBody Proj
 					@ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content),
 					@ApiResponse(responseCode = "403", description = "Forbidden", content = @Content) })
 	@GetMapping
+
```
pkliczewski, line 85: is this extra line needed?

workflow-service/src/test/java/com/redhat/parodos/workflow/definition/dto/WorkFlowPropertiesDefinitionDTOTest.java
```diff
@@ -0,0 +1,34 @@
+package com.redhat.parodos.workflow.definition.dto;
+
+import com.redhat.parodos.workflow.definition.entity.WorkFlowPropertiesDefinition;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
pkliczewski, line 6: What do you think about expanding the asterisk?

pkliczewski (commented):
minor comments, overall looks good

parodos-model-api/src/main/java/com/redhat/parodos/workflow/annotation/WorkFlowProperties.java
eloycoto: Because this is something that the user want to use on workflow-examples jar for example, and if we do here, the git  commit will be the service instead of the user jar commit!

gabriel-farache: Ah! Ok I see, makes sense

pkliczewski (commented):
/lgtm

pkliczewski:
@eloycoto please rebase. Tide is complaining :)

workflow-engine/src/main/java/com/redhat/parodos/workflows/workflow/WorkFlowPropertiesMetadata.java
RichardW98: > What about generating getter and setter?
@Data has getter and setter

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/entity/WorkFlowPropertiesDefinition.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Setter;
+
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
+public class WorkFlowPropertiesDefinition {
+
+	@Getter
+	@Setter
```
RichardW98, line 17: @Data has getter&setter, no need to add them here for each property again

workflow-service/src/main/resources/db/changelog/tables.xml
```diff
@@ -53,6 +53,7 @@
             <column name="processing_type" type="varchar(255)"/>
             <column name="number_of_works" type="int"/>
             <column name="commit_id" type="varchar(255)"/>
+            <column name="properties" type="jsonb"/>
```
RichardW98, line 56: please confirm jsonb is available in h2

workflow-service/src/main/java/com/redhat/parodos/workflow/version/WorkFlowVersionServiceImpl.java
```diff
@@ -33,6 +33,11 @@
 @Slf4j
 public class WorkFlowVersionServiceImpl implements WorkFlowVersionService {
 
+	public static String GetVersionHashForObject(Object workflowObject) throws IOException {
+		WorkFlowVersionServiceImpl versionService = new WorkFlowVersionServiceImpl();
```
RichardW98, line 37: If you want to make it a Delegate/Util,  try to make the method "getHash" static  and remove "implements WorkFlowVersionService".

RichardW98 (approved):

PR #200, opened 2023-04-03
Simple Flow with SDK

gciavarrini:
Rewrite the `run_simple_flow` function of `run_examples.sh` script.
Assertions added to ensure a proper behavior of all the invocations.

### DIFF
--- a/workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
+++ b/workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
@@ -1,0 +1,175 @@
+package com.redhat.parodos.examples.integration;
+
+import com.redhat.parodos.examples.integration.utils.ExamplesUtils;
+import com.redhat.parodos.sdk.api.ApiClient;
+import com.redhat.parodos.sdk.api.ApiException;
+import com.redhat.parodos.sdk.api.Configuration;
+import com.redhat.parodos.sdk.api.ProjectApi;
+import com.redhat.parodos.sdk.api.WorkflowApi;
+import com.redhat.parodos.sdk.api.WorkflowDefinitionApi;
+import com.redhat.parodos.sdk.model.ArgumentRequestDTO;
+import com.redhat.parodos.sdk.model.ProjectRequestDTO;
+import com.redhat.parodos.sdk.model.ProjectResponseDTO;
+import com.redhat.parodos.sdk.model.WorkFlowDefinitionResponseDTO;
+import com.redhat.parodos.sdk.model.WorkFlowRequestDTO;
+import com.redhat.parodos.sdk.model.WorkFlowResponseDTO;
+import com.redhat.parodos.sdk.model.WorkFlowResponseDTO.WorkStatusEnum;
+import com.redhat.parodos.sdk.model.WorkRequestDTO;
+import com.redhat.parodos.workflow.consts.WorkFlowConstants;
+import com.redhat.parodos.workflow.enums.WorkFlowProcessingType;
+import com.redhat.parodos.workflow.enums.WorkFlowType;
+import com.redhat.parodos.workflow.enums.WorkType;
+import com.redhat.parodos.workflow.utils.CredUtils;
+import lombok.extern.slf4j.Slf4j;
+import org.assertj.core.util.Strings;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.http.HttpHeaders;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * @author Gloria Ciavarrini (Github: gciavarrini)
+ */
+@Slf4j
+public class SimpleWorkFlow {
+
+	private static final String projectName = "project-1";
+
+	private static final String projectDescription = "an example project";
+
+	@Before
+	public void setUp() throws IOException {
+
+	}
+
+	@Test
+	public void runSimpleWorkFlow() {
+		log.info("Running simple flow");
+		ApiClient apiClient = Configuration.getDefaultApiClient();
+		apiClient.addDefaultHeader(HttpHeaders.AUTHORIZATION, "Basic " + CredUtils.getBase64Creds("test", "test"));
+
+		ProjectApi projectApi = new ProjectApi(apiClient);
+		try {
+			ExamplesUtils.waitProjectStart(projectApi);
+		}
+		catch (ApiException e) {
+			fail(e.getMessage());
+		}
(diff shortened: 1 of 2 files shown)


workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
```diff
@@ -0,0 +1,175 @@
...
+
+		ProjectResponseDTO testProject;
+
+		try {
+			// RETRIEVE ALL PROJECTS AVAILABLE
+			log.info("Get all available projects");
+			List<ProjectResponseDTO> projects = projectApi.getProjects();
+
+			// CHECK IF testProject ALREADY EXISTS
+			testProject = projects.stream()
```
pkliczewski, line 79: What do you think about extracting this to a method? It is used here and in line 103.

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
```diff
@@ -0,0 +1,175 @@
...
+			assertNotNull(workFlowResponseDTO.getWorkFlowExecutionId());
+			assertNull(workFlowResponseDTO.getWorkFlowOptions());
+			assertNotNull(workFlowResponseDTO.getWorkStatus());
+			assertEquals(WorkStatusEnum.COMPLETED, workFlowResponseDTO.getWorkStatus());
+
+			log.info("workflow finished successfully with response: {}", workFlowResponseDTO);
+			log.info("******** Simple Sequence Flow Completed ********");
+
+		}
+		catch (ApiException e) {
```
pkliczewski, line 170: Why not to use `ApiException` to fail the test? Is there a value in catching it?

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/utils/ExamplesUtils.java
```diff
@@ -0,0 +1,27 @@
...
+public final class ExamplesUtils {
+
+	public static void waitProjectStart(ProjectApi projectApi) throws ApiException {
+		ApiResponse<List<ProjectResponseDTO>> projectsWithHttpInfo = projectApi.getProjectsWithHttpInfo();
+		for (int i = 0; i < 100; i++) {
+			if (projectsWithHttpInfo.getStatusCode() == HttpStatus.SC_OK) {
+				return;
+			}
+		}
+		System.out.println("wait");
```
pkliczewski, line 23: Let's pass log to this method so we could use it here instead of system.out

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/utils/ExamplesUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.List;
+
+/**
+ * @author Gloria Ciavarrini (Github: gciavarrini)
+ */
+public final class ExamplesUtils {
+
+	public static void waitProjectStart(ProjectApi projectApi) throws ApiException {
+		ApiResponse<List<ProjectResponseDTO>> projectsWithHttpInfo = projectApi.getProjectsWithHttpInfo();
+		for (int i = 0; i < 100; i++) {
```
pkliczewski, line 18: What do you think about sleeping instead of busy waiting? Maybe there is async way of doing it?

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
gciavarrini: No value, you're right! I'll remove it

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/utils/ExamplesUtils.java
gciavarrini: Forgot to add a `sleep` inside the for loop. 

By the way, I was not sure on how to implement this. There is the possibility to use [spring-retry](https://www.baeldung.com/spring-retry) but I'm not sure if this is "too much" for an example package.

What do you think?

nirarg (commented):
This test should be ignored during the `build` ci job and to be included in the `integration` ci job, what do you say?

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/utils/ExamplesUtils.java
pkliczewski: There is a way to use `OkHttpClient` in async way. It is a client used internally by `ApiClient` and there is a `getHttpClient` method. Here is an [example](https://stackoverflow.com/questions/33418232/okhttp-update-ui-from-enqueue-callback#33427922) how to do it.

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
```diff
@@ -0,0 +1,175 @@
...
+
+	}
+
+	@Test
+	public void runSimpleWorkFlow() {
+		log.info("Running simple flow");
+		ApiClient apiClient = Configuration.getDefaultApiClient();
+		apiClient.addDefaultHeader(HttpHeaders.AUTHORIZATION, "Basic " + CredUtils.getBase64Creds("test", "test"));
+
+		ProjectApi projectApi = new ProjectApi(apiClient);
```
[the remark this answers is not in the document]
pkliczewski: I think @gciavarrini wanted to reimplement our example [script](https://github.com/parodos-dev/parodos/blob/main/workflow-examples/run_examples.sh#L134). It seems to match the flow there.

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/utils/ExamplesUtils.java
gciavarrini: Addressed. Can you please take a look? I also implemented a retry mechanism similar to the one available in the bash script but I'm not completely sure about it.

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
```diff
@@ -0,0 +1,175 @@
...
+
+	@Before
+	public void setUp() throws IOException {
+
+	}
+
+	@Test
+	public void runSimpleWorkFlow() {
+		log.info("Running simple flow");
+		ApiClient apiClient = Configuration.getDefaultApiClient();
```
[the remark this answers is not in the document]
gciavarrini: Done

workflow-examples/src/test/java/com/redhat/parodos/examples/integration/SimpleWorkFlow.java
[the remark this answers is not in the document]
gciavarrini: As @pkliczewski said, I've just reimplemented the bash script. 
If it's ok for you @nirarg, we can create some utility functions in future PR to optimize example implementations.
I'd like to stick, somehow, with the current implementation to ease the review and the migration from bash script to java.

pkliczewski (commented):
/lgtm

eloycoto:
/lgtm 

I'm ok to have it in another PR :-)

PR #202, opened 2023-04-03
Arrange modules order and dependencies

nirarg:
* Change modules order in main pom.xml to match dependencies between the modules
* Add dependency between "workflow-service" to "prebuilt-tasks"


rgolangh:
/lgtm

arguable pattern-detection-library should go in above the examples as well, but I'll leave this to anyone who will create an example using that lib

PR #206, opened 2023-04-04
Fix: enable formatter for service SDK generated files

gciavarrini:
Execution of `mvn verify` leads to formatting differences.


workflow-service-sdk/pom.xml
```diff
@@ -27,10 +28,15 @@
...
                 <groupId>io.spring.javaformat</groupId>
                 <artifactId>spring-javaformat-maven-plugin</artifactId>
-                <version>0.0.34</version>
-                <configuration>
-                    <skip>true</skip>
-                </configuration>
+                <version>${spring.javaformat.version}</version>
+                <executions>
+                    <execution>
+                        <phase>validate</phase>
```
masayag, line 34: perhaps `apply` is the better option? so we can enforce the formatting of the generated files.
unless we already instruct the generator which formatted to use

eloycoto (approved):

workflow-service-sdk/pom.xml
```diff
@@ -27,10 +28,15 @@
...
                 <groupId>io.spring.javaformat</groupId>
                 <artifactId>spring-javaformat-maven-plugin</artifactId>
-                <version>0.0.34</version>
-                <configuration>
-                    <skip>true</skip>
-                </configuration>
+                <version>${spring.javaformat.version}</version>
+                <executions>
+                    <execution>
+                        <phase>verify</phase>
```
masayag, line 34: I think the phase is wrong here.

We generate the files at the default phase.
`verify` phase is invoked after the sources are already compiled and packaged.
This means the content of the jar will differ from the actual source.

I suggest being explicit of both:
generate-sources - for generating the sources
process-resources - for formatting the sources

see https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#default-lifecycle

masayag (changes requested):

workflow-service-sdk/pom.xml
gciavarrini: Done, thanks!

PR #208, opened 2023-04-04
pom.xml: use dependecy management

rgolangh:
- declare dependecy at the parent, including version and scope
- use deps at the module level, with only the group and id
- remove conficting declerations, redundant properties

### DIFF
--- a/workflow-examples/pom.xml
+++ b/workflow-examples/pom.xml
@@ -20,17 +20,9 @@
         <url>http://github.com/parodos-dev/parodos/tree/master</url>
     </scm>
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>11</java.version>
-        <junit.version>4.13.1</junit.version>
-        <assertj.version>3.17.2</assertj.version>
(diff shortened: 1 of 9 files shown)


masayag:
Thank you for pushing this PR. I had a similar thought that we need this exact one to align project-wide dependencies.
Pls resolve conflict issues and let's merge it.

workflow-examples/pom.xml
```diff
@@ -20,17 +20,9 @@
         <url>http://github.com/parodos-dev/parodos/tree/master</url>
     </scm>
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>11</java.version>
-        <junit.version>4.13.1</junit.version>
-        <assertj.version>3.17.2</assertj.version>
-        <mockito.version>3.5.13</mockito.version>
         <slf4j.version>1.7.30</slf4j.version>
         <lombok.version>1.18.26</lombok.version>
```
masayag, line 24: just noticed, lombok.version can be pulled to parent pom as well.

PR #210, opened 2023-04-04
ci: Use mvn verify to process the generate files as well

rgolangh:
This fixes the constant failure of `mvn clean spring-javaformat:validate
package` because the generate files where not formatted yet

The default phase of the formatter is on validate phase which is too
early. mvn verify is enough here.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -17,8 +17,8 @@ jobs:
           distribution: 'temurin'
           java-version: '11'
 
-      - name: Maven Build
-        run: mvn clean spring-javaformat:validate package
+      - name: Maven Verify 
+        run: mvn verify 
 
       - name: Check generated files are up to date
         run: |
--- a/notification-service-sdk/pom.xml
+++ b/notification-service-sdk/pom.xml
@@ -184,48 +184,21 @@
                     </execution>
                 </executions>
             </plugin>
-            <!-- Use spotless plugin to automatically format code, remove unused import, etc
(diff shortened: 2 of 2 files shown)


.github/workflows/push.yml
```diff
@@ -17,8 +17,8 @@ jobs:
           distribution: 'temurin'
           java-version: '11'
 
-      - name: Maven Build
-        run: mvn clean spring-javaformat:validate package
```
masayag, line 21: `spring-javaformat:validate` is designed for all code, not only for the generated one.
With this change, the CI will not enforce code updates (non-generated) format.

rgolangh: i.e on the test.yml

still this is not good that validate misses formatting on regular code... looking further

masayag: well, it was named "Maven Build" before you changed it :-) 
This is the first job that is run when a PR is pushed.

rgolangh: just to make sure I'm not wrong - on:pull_request should fire the test action . on:push should fire the push action

the test action should fail if we missed the formatting, of all code
the push should add a sanity if we want... not sure about that yet

masayag: So here is my confusion:
This PR runs when a PR is merged.
I'm happy with changing this to `mvn verify` but we need to make sure `spring-javaformat:validate` is part of parent pom plugin.
I also think that we should preserve `clean` goal here. Not sure about the state of the FS at this point.

PR #211, opened 2023-04-04
CI: Rebase code with main branch when running the action

eloycoto:
On PR rebase the current code with the main branch, if there is any
    conflict will fail, but code will run against the target branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,16 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha != '' &&  github.event.pull_request.head.sha ||  github.sha }}
 
+      - name: Fetch target branch and rebase it
+        if: github.event_name == 'pull_request_target'
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -19,6 +19,16 @@ jobs:
...
+      - name: Fetch target branch and rebase it
+        if: github.event_name == 'pull_request_target'
+        run: |
+          set -x
+          git config user.email "CI@action.git"
+          git config user.name "CI"
+          git remote add upstream https://github.com/${{ github.repository }}.git
+          git fetch upstream ${{ github.event.pull_request.base.ref }}
+          git rebase upstream/${{ github.event.pull_request.base.ref }}
+
```
masayag, line 31: I'm surprised there isn't action ready for that. I'd imagine this should be pretty common.

eloycoto: not at all, there is the pull_request_target event..

PR #214, opened 2023-04-04
fix getting workflow parameter

RichardW98:
fix getting workflow parameter

### DIFF
--- a/workflow-service/src/main/java/com/redhat/parodos/workflow/WorkFlowDelegate.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/workflow/WorkFlowDelegate.java
@@ -16,6 +16,8 @@
 package com.redhat.parodos.workflow;
 
 import com.redhat.parodos.workflow.context.WorkContextDelegate;
+import com.redhat.parodos.workflow.definition.dto.WorkDefinitionResponseDTO;
+import com.redhat.parodos.workflow.definition.dto.WorkFlowDefinitionResponseDTO;
 import com.redhat.parodos.workflow.enums.WorkType;
 import com.redhat.parodos.workflow.execution.dto.WorkFlowRequestDTO;
 import com.redhat.parodos.workflow.registry.BeanWorkFlowRegistryImpl;
@@ -24,6 +26,9 @@ import com.redhat.parodos.workflows.work.WorkContext;
 import com.redhat.parodos.workflows.workflow.WorkFlow;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
+
+import java.util.List;
+import java.util.Optional;
 
 /**
  * Provides functionality that is common to any WorkFlow composition in Parodos
@@ -43,7 +48,8 @@ public class WorkFlowDelegate {
 		this.beanWorkFlowRegistry = beanWorkFlowRegistry;
 	}
 
-	public WorkContext initWorkFlowContext(WorkFlowRequestDTO workFlowRequestDTO) {
+	public WorkContext initWorkFlowContext(WorkFlowRequestDTO workFlowRequestDTO,
+			WorkFlowDefinitionResponseDTO masterWorkFlowDefinitionDto) {
 		WorkContext workContext = new WorkContext();
 
 		if (workFlowRequestDTO.getArguments() != null && !workFlowRequestDTO.getArguments().isEmpty()) {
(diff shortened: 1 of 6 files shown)


workflow-service/src/main/java/com/redhat/parodos/workflow/WorkFlowDelegate.java
```diff
@@ -43,34 +45,57 @@ public WorkFlowDelegate(BeanWorkFlowRegistryImpl beanWorkFlowRegistry) {
...
+		if (workFlowRequestDTO != null && workFlowRequestDTO.getWorks() != null)
+			for (WorkFlowRequestDTO.WorkRequestDTO workRequestDTO : workFlowRequestDTO.getWorks())
+				if (name.equals(workRequestDTO.getWorkName()))
+					return workRequestDTO;
+		return null;
+	}
+
+	private WorkFlowRequestDTO.WorkRequestDTO findSubWorkFromRequest(WorkFlowRequestDTO.WorkRequestDTO workRequestDTO,
+			String name) {
+		if (workRequestDTO != null && workRequestDTO.getWorks() != null)
```
eloycoto, line 94: I would change this to make it simple, like 

if (workRequestDTO == null || workRequestDTO.getWorks() == null) return null

It's more cognitive based

workflow-service/src/main/java/com/redhat/parodos/workflow/WorkFlowDelegate.java
```diff
@@ -43,34 +45,57 @@ public WorkFlowDelegate(BeanWorkFlowRegistryImpl beanWorkFlowRegistry) {
...
+			for (WorkFlowRequestDTO.WorkRequestDTO subWorkRequestDTO : workRequestDTO.getWorks())
```
eloycoto, line 95: shouldn't we use here `stream.filter()`?

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceImpl.java
```diff
@@ -158,10 +159,11 @@ public WorkFlowDefinitionResponseDTO save(String workFlowName, WorkFlowType work
 									.workFlowDefinition(finalWorkFlowDefinition).createDate(new Date())
 									.modifyDate(new Date()).build()))
 							.getId();
-					workFlowWorkDefinitions.add(
-							WorkFlowWorkDefinition.builder().workDefinitionId(workId).workDefinitionType(WorkType.TASK)
-									.workFlowDefinition(finalWorkFlowDefinition).createDate(new Date()).build());
 				}
+				workFlowWorkDefinitions.add(WorkFlowWorkDefinition.builder()
+						.workDefinitionId(workFlowTaskDefinitionOptional.map(AbstractEntity::getId).orElse(workId))
```
eloycoto, line 164: Could you explain why we use the definitionOptional? I'm not sure about it!

workflow-service/src/test/java/com/redhat/parodos/workflow/WorkFlowDelegateTest.java
```diff
@@ -87,9 +91,19 @@ class WorkFlowDelegateTest {
 					.build()))
 			.build();
 
+	private WorkFlowDefinitionResponseDTO workFlowDefinitionResponseDTO = WorkFlowDefinitionResponseDTO
```
eloycoto, line 94: I would move this to a, sampleWorkFlowDEfintinoREsponse() function, so if any other test modified this, it will not be flake.

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
```diff
@@ -101,7 +106,9 @@ public WorkReport execute(WorkFlowRequestDTO workFlowRequestDTO) {
 			return new DefaultWorkReport(WorkStatus.FAILED, new WorkContext(), new Throwable(validationFailedMsg));
 		}
 
-		WorkContext workContext = workFlowDelegate.initWorkFlowContext(workFlowRequestDTO);
+		WorkFlowDefinition workFlowDefinition = workFlowDefinitionRepository.findFirstByName(workflowName);
```
eloycoto, line 109: At some point, I think that we need to get rid of workflowDefinitionRepository from here, Does not make sense that from another domain we keep using internal definitions instead of DTO

eloycoto (changes requested):

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
RichardW98: agree, I will change to getByName

RichardW98: I think we should build the work relations from internal and not rely on the payload from UI.

PR #215, opened 2023-04-05
Prometheus metric

eloycoto:
Hi, 

Three commits mainly regarding this:

- Two that enable metrics on both spring boot projects. 
- One that adds custom metrics on workflow-executions progress changes.


workflow-service/src/main/resources/application-local.yml
```diff
@@ -20,16 +20,3 @@ spring:
 springdoc:
   writer-with-order-by-keys: true
   writer-with-default-pretty-printer: true
-management:
```
pkliczewski, line 23: Why do we want to remove this here and for dev profile?

notification-service/pom.xml
```diff
@@ -185,6 +186,14 @@
             <artifactId>jackson-databind-nullable</artifactId>
             <version>${jackson-databind-nullable-version}</version>
         </dependency>
+        <!-- START Metrics Dependencies -->
```
pkliczewski, line 189: Do you think it makes sense to follow https://github.com/parodos-dev/parodos/pull/208

workflow-service/src/main/resources/application-local.yml
eloycoto: It's already part of the application.yaml, and it's not needed at all.

notification-service/pom.xml
eloycoto: not really, the thing is only used on workflow-service and notification-service. so I do not see a need.

RichardW98 (approved):

PR #217, opened 2023-04-05
Refactor: small refactor on workflow-definition-service

eloycoto:


workflow-service/src/main/java/com/redhat/parodos/workflow/definition/service/WorkFlowDefinitionServiceImpl.java
```diff
@@ -106,30 +106,26 @@ public WorkFlowDefinitionResponseDTO save(String workFlowName, WorkFlowType work
...
 		}
+
+		workFlowDefinition.setType(workFlowType);
+		workFlowDefinition.setParameters(stringifyParameters);
+		workFlowDefinition.setModifyDate(new Date());
+		workFlowDefinition.setProperties(propertiesDefinition);
+		workFlowDefinition.setProcessingType(workFlowProcessingType.name());
+		workFlowDefinition.setNumberOfWorks(works.size());
+
+		workFlowDefinition = workFlowDefinitionRepository.save(workFlowDefinition);
```
RichardW98, line 127: we only want to save the workflowDefinition if it's not found in db or updated.

eloycoto: The thing is, does it matter in terms of complexity? At the moment, if the works change inside, but keep the same number will never be updated. 

Maybe, what we can do, it's avoid that save() and only use the last one (finally) something that I'm happy to do

RichardW98: no I think this is good,   let's update it every time when app restarts

RichardW98 (approved):

PR #233, opened 2023-04-13
ci: add clean before verify

rgolangh:
Bringing back 'clean' before the verify as we have
1. some pluging relying on that (like in external-dependencies module)
2. this makes sure we verify on a clean state in the running env

note - clean is not part of the verify lifecycle and maybe downloading
jar in external deps doesn't make sense in clean phase. dependency
resolution like this make more sense in 'compile'.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -18,7 +18,7 @@ jobs:
           java-version: '11'
 
       - name: Maven Verify 
-        run: mvn verify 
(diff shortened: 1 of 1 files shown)


rgolangh:
annoying glitch in run examples... didn't look into why it failed yet

rgolangh:
now lets hope after merge the post action will be :green_heart:

PR #234, opened 2023-04-13
workflow-service-sdk version changes

nirarg:


pkliczewski (approved):

rgolangh:
/lgtm
I think this readme itself should be generated just as well

rgolangh:
alternatively use a range syntax like `1.0` and let it pick what's needed. 

if the autogeneration already works, we are then missing the CI to tell us something has changed. I'm planning on adding this

gciavarrini:
@rgolangh  FYI [here](https://github.com/parodos-dev/parodos/blob/429d7e6e7ec378580c0d771fabdae8530d0501f5/.github/workflows/push.yml#L23-L26) there is a CI step that checks if something changed. Maybe it is not working properly.

PR #235, opened 2023-04-13
Migrate to Java 17

rgolangh:
- No code changes
- pom.xml changes for maven compiler plugin

[FLPATH-296](https://issues.redhat.com//browse/FLPATH-296)
Depends-on: #237 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,9 @@ TESTDBUSER := parodos
 TESTDBPORT := 5432
 
 # maven
-MAVEN ?= mvn
+MAVEN ?= /usr/bin/mvn
 
 # set quiet mode by default
-Q=@
 
 ORG=quay.io/parodos/
 WORKFLOW_SERVICE_IMAGE=workflow-service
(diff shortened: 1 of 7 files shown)


rgolangh:
extracted the CI change to #237 and once it is merged I'll rebase

rgolangh:
/hold
the makefile changes broke the java-checks target

workflow-service/src/test/java/com/redhat/parodos/workflow/version/WorkFlowVersionServiceImplTest.java
```diff
@@ -19,13 +17,17 @@ public void setUp() {
...
+		// consistent, because java
+		// can change something in the implementation (or even compile flags) that will
+		// cause generation of a different
+		// digest, so we can't test reliably against a constant digest.
+		// If we want a stable test we probably need to change the argument of getHash
+		// from Object to some InputStream
 		Object object = new Object();
 
 		// when
-		assertDoesNotThrow(() -> {
```
sergiubodiu, line 25: This is not the best approach to check the value, if we maintain the hardcoded hash for compatibility purposes because the workflows will be versioned in the database, we need to have a migration test.
If we go ahead with this change, the assertion test doesn't make sense and we should remove this test all together

rgolangh: I left a comment to try to describe the problem. I agree this is not the best approach, however the test as it was has a flaw that will make it fail without a code change. 
The test checks the minimum it can so I don't think we should remove it, but instead fix it with implementation -  getHash on class doesn't guarantee a class change detection because simply changin a JVM even minor version would result a difference in the hash @RichardW98

rgolangh:
/unhold
The `mvn-checks` and `java-checks` where missing tabs so they where not really targets, so they always got invoked, and with this change it suddenly failed because the CI actions are having mismatch on the version. In the `build-images` the and the rest of the docker build they aren't needed and will not get invoked - this is fixed now.

PR #248, opened 2023-04-17
rename master workflow to main

anludke:


pkliczewski:
I noticed one [place](https://github.com/parodos-dev/parodos/blob/main/workflow-examples/src/main/java/com/redhat/parodos/examples/ocponboarding/task/OcpAppDeploymentWorkFlowTask.java#L73) where you missed rename. Please fix in new PR

PR #249, opened 2023-04-18
Changing the history to fix tags

pkliczewski:
Recently we started rebasing commits which is causing commit hashes to change. This caused tags to be out of the tree.
This commit fixes the issue.

No file changes. Only moved tags which changed the history.


pkliczewski:
I seems this PR won't be rebased

PR #253, opened 2023-04-18
Actions: refactor

eloycoto:
- Report the coverage report using the GITHUB_STEPS
    - Use github_pull_request instead of target.(No longer need the target)
    - Use the merge commit instead of the PR one.
    - Fail if coverage fails down from 65%
    - Build no longer run tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
       - ActionsTest
       - 'releases/**'
 
-  pull_request_target:
+  pull_request:
     branches:
       - '**'
 permissions: read-all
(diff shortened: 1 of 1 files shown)


pkliczewski (commented):
/lgtm

if possible I would compare with previous coverage report and failed if it was lower.

masayag:
> /lgtm
> 
> if possible I would compare with previous coverage report and failed if it was lower.

It is below the current bar, but pretty close to it:
https://github.com/eloycoto/parodos/actions/runs/4731957559
Depends if future PRs will lower the bar before it will get merged...

PR #255, opened 2023-04-18
entity: Use Enum WorkflowProcessingType instead of String

rgolangh:
Strong type is preferable where available. Prevents error (like the
uncought one in the test) and more readable code.

JPA performs the conversion for us, and there are no schema
changes.


eloycoto (commented):
Thanks for doing this :heart: 

/approve
/lgtm

RichardW98 (approved):

PR #266, opened 2023-04-19
Authentication: No Authentication when using oauth-proxy

eloycoto:
Because we choose to use oauth-proxy in front, when openshift is used, we need to disable all the Authentication due to is made in the proxy.


pkliczewski (commented):
@eloycoto do you think it makes sense to add an integration test?

eloycoto:
I already said, having an integration test for this is going to be complicated. We need an entire Openshift instance to be able to test this, and I'm not aware of anything like this in the GH actions.

RichardW98 (approved):

PR #268, opened 2023-04-20
Add Pull Request template

nirarg:
I open this PR as suggestion and discussion about what should be our guidelines for Pull Requests

The purpose of this template is to provide the author with a clearer set of instructions on the necessary steps to follow in order to create a proper pull request. As a result, the review process will be streamlined and more effective.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,29 @@
+<!--
+- Please ensure code changes are split into a series of logically independent commits.
+- Every commit should have a subject/title (What) and a description/body (Why).
+- Every PR must have a description.
+- As an example you can use git commit -m"What" -m"Why" to achieve the requirements above. GitHub automatically recognises the commit description (-m"Why") in single commit PRs and adds it as the PR description.
+- If the PR includes auto generated code, make sure to add this code in different commit (last commit), to make the review process easier.
+-->
+
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story (FLPATH-xxxx)*:
+Fixes #
+
+**Change type**
+- [ ] New feature
+- [ ] Bug fix
+- [ ] Unit tests
+- [ ] Integration tests
+- [ ] CI
+- [ ] Documentation
+- [ ] Auto generated SDK code
+
+**Impacted services**
+- [ ] Workflow Service
+- [ ] Notifivcation Service
+
+**Checklist**
+- [ ] Subject and description added to both, commit and PR.
+- [ ] Relevant issues have been referenced.


eloycoto:
I'm not sure if this helps at all, my experience is:
- At the moment, we don't even force good commits. So, IMHO, we're far away from writing good PR descriptions.
- The templates, in my experience, only help to have a placeholder that no one uses.

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
+Fixes #
+
+**Change type**
+- [ ] New feature
+- [ ] Bug fix
+
+**Impacted elements**
```
masayag, line 18: I'd merge this one with the above

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+- [ ] Unit tests
+- [ ] Integration tests
+- [ ] CI
+- [ ] Docs
```
masayag, line 22: s/Docs/Documentation

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Impacted services**
+- [ ] Workflow Service
+- [ ] Notifivcation Service
+
+**Checklist**
+- [ ] Subject and description added to both, commit and PR.
+- [ ] Relevant issues have been referenced.
+- [ ] `mvn clean install` was executed to validate changes
+- [ ] Integration tests were validated with the change
```
masayag, line 33: These two will be run by the CI, I don't think there is a need to check them manually

nirarg:
> I'm not sure if this helps at all, my experience is:
> 
> * At the moment, we don't even force good commits. So, IMHO, we're far away from writing good PR descriptions.
> * The templates, in my experience, only help to have a placeholder that no one uses.

I think this is a good thing to start with
And this is our responsibility as a group to enforce this (if we think it may help)

.github/pull_request_template.md
nirarg: But it can be good reminder for the Author, before the PR is opened

masayag: it is implied by `**Impacted elements**` section

rgolangh:
I would take this and put it as the start of CONTRIBUTING.md and consider making this template bit more concise.
Doesn't have to be in this PR because it is a good improvement from what we have now.

rgolangh:
> > I'm not sure if this helps at all, my experience is:
> > 
> > * At the moment, we don't even force good commits. So, IMHO, we're far away from writing good PR descriptions.
> > * The templates, in my experience, only help to have a placeholder that no one uses.
> 
> I think this is a good thing to start with And this is our responsibility as a group to enforce this (if we think it may help)



> > I'm not sure if this helps at all, my experience is:
> > 
> > * At the moment, we don't even force good commits. So, IMHO, we're far away from writing good PR descriptions.
> > * The templates, in my experience, only help to have a placeholder that no one uses.
> 
> I think this is a good thing to start with And this is our responsibility as a group to enforce this (if we think it may help)

in time if this is just a burden we can narrow this down

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+- Please ensure code changes are split into a series of logically independent commits.
+- Every commit should have a subject/title (What) and a description/body (Why).
+- Every PR must have a description.
+- As an example you can use git commit -m"What" -m"Why" to achieve the requirements above. GitHub automatically recognises the commit description (-m"Why") in single commit PRs and adds it as the PR description.
+- If the PR includes auto generated code, make sure to add this code in different commit (last commit), to make the review process easier.
+-->
+
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
```
rgolangh, line 11: Here we can put "FLPATH-xxxx  " as an example to what we expect

rgolangh:
@nirarg since you suggested it and there's some debate around it - can you take revisit this with the team in say 2 weeks, 1 month from now?

.github/pull_request_template.md
nirarg: As I see it, `**Impacted elements**` indicates if the element code was changed (i.e. changed integration test)
But even if the integration code is not changed, it need to be validated

Any way, I'm flexible on this matter and it can be eliminated to simplify the template.

pkliczewski (commented):
[Here](https://github.com/openshift/assisted-service/blob/master/docs/pull_request_template.md) is an example how assisted installer team defined a template.

.github/pull_request_template.md
nirarg: Done

.github/pull_request_template.md
nirarg: Done

.github/pull_request_template.md
nirarg: Removed those lines

pkliczewski (commented):
/lgtm

masayag:
/lgtm

not yet approving to allow others to add their comments.

PR #272, opened 2023-04-23
Add a workflow task that manipulates kubernetes resources

ydayagi:
[FLPATH-144](https://issues.redhat.com//browse/FLPATH-144) https://issues.redhat.com/browse/FLPATH-144


prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,171 @@
...
+					break;
+				case CREATE:
+					create(workContext, apiGroup, apiVersion, kindPluralName);
+					break;
+				case GET:
+					get(workContext, apiGroup, apiVersion, kindPluralName);
+					break;
+			}
+		}
+		catch (Exception e) {
```
pkliczewski, line 116: What exception do we expecting here? Would it make sense to be more specific?

workflow-examples/run_kubeapi.sh
```diff
@@ -0,0 +1,243 @@
...
+                  "key": "resource-json",
+                  "value": "{\"apiVersion\": \"v1\",\"data\": {\"a\": \"valb\",\"b\": \"vala\"},\"kind\": \"ConfigMap\",\"metadata\": {\"name\": \"my-cm-create\",\"namespace\": \"flpath144\"}}"
+                }
+            ]
+          }
+        ]
+      }' | jq -r '.workFlowExecutionId')"
+}
+
+run_flow() {
```
pkliczewski, line 177: what do you think about using the sdk? we are moving away from bash based tests

workflow-service/pom.xml
```diff
@@ -36,7 +36,11 @@
     </properties>
 
     <dependencies>
-        <!-- START Web Dependencies -->
+        <dependency>
+            <groupId>io.kubernetes</groupId>
```
pkliczewski, line 40: Why do we need this update here?

pkliczewski (commented):
The test are failing due to:

> org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.redhat.parodos.tasks.kubeapi.KubeapiWorkFlowTask]: Factory method 'kubeapiTask' threw exception; nested exception is java.nio.file.NoSuchFileException: kubeconfig

please fix

workflow-examples/run_kubeapi.sh
ydayagi: not in this PR. We can open a separate story for that

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
ydayagi: no. This method is expected not to throw. specifying an exception type is only useful when u have different handling for different types.

workflow-examples/run_kubeapi.sh
ydayagi: BTW piotr, this is not a test. it is an example.

workflow-service/pom.xml
ydayagi: we have a few dependencies that depend on this artifact. Each one wants a different version. The way to solve it is by declaring the dependency directly in the higher level pom so it forces a version for all of them.  Obviously, the highest version since it is backward compatible.

workflow-examples/run_kubeapi.sh
eloycoto: Then, add the test in a different commit with the test.

ydayagi:
>

ydayagi:
> The test are failing due to:
> 
> > org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.redhat.parodos.tasks.kubeapi.KubeapiWorkFlowTask]: Factory method 'kubeapiTask' threw exception; nested exception is java.nio.file.NoSuchFileException: kubeconfig
> 
> please fix

fixed

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
pkliczewski: What I mean is to be more explicit about what we catch here so it is easier to understand.

We handle 3 exceptions:

> ApiException | MissingParameterException | IOException

workflow-examples/run_kubeapi.sh
pkliczewski: we are moving away from bash script. please follow the new way

workflow-service/pom.xml
pkliczewski: you are right, I see issue with versions and we need to be explicit

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTaskTest.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Test
+	public void get() throws ApiException {
+		String objJson = "{\"kind\":\"ConfigMap\",\"apiVersion\":\"v1\",\"metadata\":{\"name\":\"my-cm\",\"namespace\":"
+				+ "\"test\",\"uid\":\"5f28eebc-05b2-41b0-9da1-e7399adc1262\",\"resourceVersion\":\"163811\","
+				+ "\"creationTimestamp\":\"2023-04-16T18:47:25Z\"},\"data\":{\"a\":\"vala\",\"b\":\"valb\"}}";
+		DynamicKubernetesObject obj = Dynamics.newFromJson(objJson);
+		WorkContext ctx = workContextGET();
+		doReturn(obj).when(api).get(any(), any(), any(), any(), any(), any());
```
jordigilh, line 55: Please use the parameters referenced to the configmap that are expected for the call instead of `any()`.
In this case you can use the `apiGroup, apiVersion, kindPluralName,namespace and name` from the manifest itself.

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTaskTest.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Test
+	public void update() throws ApiException {
+		DynamicKubernetesObject expectedObj = Dynamics.newFromJson(
+				"{\"apiVersion\": \"v1\",\"data\": {\"a\": \"valb\",\"b\": \"vala\"},\"kind\": \"ConfigMap\",\"metadata\": {\"resourceVersion\": \"1000\", \"name\": \"my-cm-create\",\"namespace\": \"test\"}}");
+		DynamicKubernetesObject getReturnValue = Dynamics.newFromJson(
+				"{\"apiVersion\": \"v1\",\"data\": {\"a\": \"vala\",\"b\": \"valb\"},\"kind\": \"ConfigMap\",\"metadata\": {\"resourceVersion\": \"1000\", \"name\": \"my-cm-create\",\"namespace\": \"test\"}}");
+		WorkContext ctx = workContextUPDATE();
+		doReturn(getReturnValue).when(api).get(any(), any(), any(), any(), any(), any());
```
jordigilh, line 78: same here.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
ydayagi: I don't see how is it easier. I personally think it is wrong. Because we want to prevent ANY exception from being thrown regardless of the code inside the try block. So the code starts with 'catch Exception' and never needs to be updated. But I'll change it.

workflow-examples/run_kubeapi.sh
ydayagi: yes, but I have a PR already. I don't want to rebase all the time. opened a story https://issues.redhat.com/browse/FLPATH-337 in the epic I am working on

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTaskTest.java
ydayagi: we do it in other tests as well. Actually, Eloy suggested I use it in my previous assignment. It makes the test coe much easier to manage.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
ydayagi: fixed

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTaskTest.java
ydayagi: we do it in other tests as well. Actually, Eloy suggested I use it in my previous assignment. It makes the test code much easier to manage.

workflow-examples/run_kubeapi.sh
ydayagi: BTW, has anyone used it for the examples?

ydayagi: removed the bash script

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTaskTest.java
jordigilh: Here you are asserting that the parameters that you pass to this method will match certain values. By specifying those values in the mock, you ensure that they are what you expect and not something else, such as empty, null or even passed in the wrong order. 
By asserting for `any()` you miss the opportunity to validate this scenario and makes the test less trustworthy.

ydayagi: I am not trying to assert/validate these values. there's no point using anything but any()

pkliczewski (commented):
overall, looks good to me. Please work on Jordi's comments

prebuilt-tasks/src/test/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTaskTest.java
jordigilh: I see. So it's on purpose. Please add the validations as I mentioned, otherwise this test is not complete.

workflow-examples/src/main/java/com/redhat/parodos/examples/KubeapiWorkFlowConfiguration.java
jordigilh, line 1: Please create a directory inside the examples directory to host this file.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,171 @@
...
+			return;
+		}
+		createClientInstance();
+	}
+
+	private synchronized void createClientInstance() throws IOException {
+		if (apiClient != null) {
+			return;
+		}
+		this.apiClient = ClientBuilder.kubeconfig(KubeConfig.loadKubeConfig(new StringReader(kubeconfig))).build();
```
jordigilh, line 168: @pkliczewski is this workflow type expected to use a [user account or a service account](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#user-accounts-versus-service-accounts) to interact with the cluster? For application processes, it's recommended to use a service account as it makes it easier to automate and has less security risks compared to using a kubeconfig derived from a user account.

Also, do we plan to allow Parodos kube tasks to reach out to other clusters? I suspect that in general  we should first check for a token + ca + url in the parameters when we want to interact remotely with a different cluster, and if not available default to retrieving the credentials from the same cluster where Parodos is running:
1) load $KUBECONFIG
2) load from $HOME/.kube/config
3) load from in-cluster service account if available

Like it's being done [here](https://github.com/kubernetes-client/java/blob/7c00056d11b80aa94fe7ccb711db3ea4f1abf06d/util/src/main/java/io/kubernetes/client/util/Config.java#L102-L110).

ydayagi: this task is for interacting with a cluster anywhere in the world. not the cluster where parodos is running. after all, parodos does not have to be in a cluster. allowing to specify a kubeconfig location enables the user to have multiple tasks with different kubeconfigs. Even for the same cluster.

ydayagi: how the kubeconfig is configured (service account, user account, token password etc) is 100 up to the user and we do not limit.

workflow-examples/src/main/java/com/redhat/parodos/examples/KubeapiWorkFlowConfiguration.java
ydayagi: the directory structure has to match the package structure. and in general the less directories the better. less confusion

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,171 @@
...
+				case CREATE:
+					create(workContext, apiGroup, apiVersion, kindPluralName);
+					break;
+				case GET:
+					get(workContext, apiGroup, apiVersion, kindPluralName);
+					break;
+			}
+		}
+		catch (MissingParameterException | IOException | ApiException | IllegalArgumentException e) {
+			log.error("Kubeapi task failed", e);
```
nirarg, line 117: Add `operationType` to the log, to have more info

workflow-examples/src/main/java/com/redhat/parodos/examples/KubeapiWorkFlowConfiguration.java
nirarg, line 1: Can you add this as new task in `PrebuiltWorkFlowConfiguration` instead of creating new workflow configuration?
I added `PrebuiltWorkFlowConfiguration` in order to include there all the pre built tasks, instead of creating new file for each task
This way, you also can use `runPreBuiltWorkFlow` integration test for testing it

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,171 @@
...
+
+	};
+
+	private final String kubeconfig;
+
+	private ApiClient apiClient;
+
+	private KubernetesApi api;
+
+	public KubeapiWorkFlowTask(String kubeconfig) {
```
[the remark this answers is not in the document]
ydayagi: there can be multiple instances of kubeapiTask. each with its own kubeconfig. therefore the problem u describe is solved. However, if each execution has its own kubeconfig then each execution performs new connection to the cluster. Performance wise it is a waste.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
ydayagi: if we add the operation type then why not add all extra details? what will you do with the operation alone?

workflow-examples/src/main/java/com/redhat/parodos/examples/KubeapiWorkFlowConfiguration.java
ydayagi: I have an epic for all prebuilt tasks. We will handle how to organize it in that epic. In general, I am in favor of a class per task. I think it is more consistent with our code. we have a class for every individual entity. I think that the file will get big and too complex to manage.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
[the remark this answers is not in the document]
ydayagi: if the need arises we can allow passing a context name.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,171 @@
...
+		String resourceName = newObj.getMetadata().getName();
+		DynamicKubernetesObject currObj = api.get(apiClient, apiGroup, apiVersion, kindPluralName, resourceNamespace,
+				resourceName);
+		V1ObjectMeta metadata = newObj.getMetadata();
+		metadata.setResourceVersion(currObj.getMetadata().getResourceVersion());
+		newObj.setMetadata(metadata);
+		api.update(apiClient, apiGroup, apiVersion, kindPluralName, newObj);
+	}
+
+	private void initClient() throws IOException {
```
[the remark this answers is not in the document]
ydayagi: instead of using a mutex directly, I used a synchronized method. This is why it is broken into two methods.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,171 @@
...
+		this.kubeconfig = "";
+		this.apiClient = apiClient;
+		this.api = api;
+		this.setBeanName(beanName);
+	}
+
+	@Override
+	public @NonNull List<WorkParameter> getWorkFlowTaskParameters() {
+		LinkedList<WorkParameter> params = new LinkedList<>();
+		params.add(WorkParameter.builder().key("api-group").type(WorkParameterType.TEXT).optional(false)
```
jordigilh, line 64: Make these as parameter names as constants so that they don't need to be redefined for each workflow that uses this one.

ydayagi: it is not per workflow. it is per workflow task. Therefore it is defined just once.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
nirarg: This function wrap Create, Update and Get operations, I think at least need to make it clear which operation (Maybe also need to consider even other parameters here)

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/kubeapi/KubeapiWorkFlowTask.java
```diff
@@ -0,0 +1,149 @@
...
+	KubeapiWorkFlowTask(KubernetesApi api, String beanName) {
+		this.api = api;
+		this.setBeanName(beanName);
+	}
+
+	@Override
+	public @NonNull List<WorkParameter> getWorkFlowTaskParameters() {
+		LinkedList<WorkParameter> params = new LinkedList<>();
+		params.add(WorkParameter.builder().key("kubeconfig-json").type(WorkParameterType.TEXT).optional(false)
+				.description("API group of resource").build());
```
nirarg, line 55: Wrong description

PR #273, opened 2023-04-23
Flpath 328: assessment workflow options should be available to the user

masayag:
The patch introduces a new API endpoint. The purpose of the new API is
to enable users to retrieve context parameters of an executed workflow,
specifically of type "assessment".

The patch adds a new controller method that maps to a new endpoint.
This new endpoint accepts the workflow execution ID and a list of context
parameters to retrieve. The method then retrieves the specified context
parameters from the workflow service and returns them as a JSON response.

An example of the request URL:
```
http://localhost:8080/api/v1/workflows/${workflowExecutionId}/context?param=WORKFLOW_OPTIONS
````

In the future, when additional parameters will be defined as 'public' visible, the URLcan be used as:
```
http://localhost:8080/api/v1/workflows/${workflowExecutionId}/context?param=WORKFLOW_OPTIONS,OTHER_PARAM
```
or a list of separate 'param's:
```
http://localhost:8080/api/v1/workflows/${workflowExecutionId}/context?param=WORKFLOW_OPTIONS&param=OTHER_PARAM
```
The response body includes the workflow execution ID and the retrieved
context parameters. An example of the response body:
```
{
  "workFlowExecutionId": "91811189-a70f-4bce-bd0b-6aa41e36986d",
  "workFlowOptions": {
    "newOptions": [
      {
        "identifier": "onboardingOption",
        "displayName": "Onboarding",
        "description": "An example of a complex WorkFlow",
        "details": [
          "An example workflow option of a complex workFlow with status checks"
        ],
        "workFlowName": "complexWorkFlow"
      }
    ]
  }
}
```
To retrieve context parameters, users will have to specify the parameters
they want to retrieve as defined in the WorkContextDelegate.Resource, for
resources that are defined as "public" visible. Currently, only the
WORKFLOW_OPTIONS context parameter is defined as "public" visible.

The schema change can be viewed in the generated openapi.json within the commit.

### CODE REVIEW

File: parodos-model-api/src/main/java/com/redhat/parodos/workflow/context/WorkContextDelegate.java [hunk 1]
```diff
@@ -15,6 +15,8 @@
  */
 package com.redhat.parodos.workflow.context;
 
+import static com.redhat.parodos.workflow.context.Visibility.*;
```
pkliczewski, line 18: can we expand the asterisk?

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
```diff
@@ -194,6 +199,33 @@ public WorkFlowStatusResponseDTO getWorkFlowStatus(UUID workFlowExecutionId) {
...
 	}
 
+	@Override
+	public WorkFlowContextResponseDTO getWorkflowParameters(UUID workFlowExecutionId,
+			List<WorkContextDelegate.Resource> params) {
+		WorkFlowExecution workFlowExecution = workFlowRepository.findById(workFlowExecutionId).orElseThrow(() -> {
+			throw new ResponseStatusException(HttpStatus.NOT_FOUND,
+					String.format("workflow execution id: %s not found!", workFlowExecutionId));
+		});
+		Map options = null;
```
pkliczewski, line 209: Is there a way to limit sensitive information leak?

parodos-model-api/src/main/java/com/redhat/parodos/workflow/context/WorkContextDelegate.java
masayag: sure, I'm surprised that happened with the IDE formatter.

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
masayag: Pls create a new issue to track this one, since this issue may exist with or without this change.

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java
```diff
@@ -112,4 +124,32 @@ public ResponseEntity<WorkFlowStatusResponseDTO> getStatus(@PathVariable String
...
+
+	@ExceptionHandler(ConstraintViolationException.class)
+	public ResponseEntity handle(ConstraintViolationException constraintViolationException) {
+		Set<ConstraintViolation<?>> violations = constraintViolationException.getConstraintViolations();
+		String errorMessage;
+		if (!violations.isEmpty()) {
+			errorMessage = violations.stream().map(ConstraintViolation::getMessage).collect(Collectors.joining(" "));
+		}
+		else {
+			errorMessage = "ConstraintViolationException occurred.";
```
pkliczewski, line 150: why not to assign this value in line 145 and get rid of else block?

masayag:
Integration tests will follow...

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java
```diff
@@ -112,4 +123,29 @@ public ResponseEntity<WorkFlowStatusResponseDTO> getStatus(@PathVariable String
...
+	@GetMapping("/{workFlowExecutionId}/context")
+	public ResponseEntity<WorkFlowContextResponseDTO> getWorkflowParameters(@PathVariable String workFlowExecutionId,
+			@NotEmpty @RequestParam List<WorkContextDelegate.@PubliclyVisible Resource> param) {
+		WorkFlowContextResponseDTO responseDTO = workFlowService
+				.getWorkflowParameters(UUID.fromString(workFlowExecutionId), param);
+		return ResponseEntity.ok(responseDTO);
+	}
+
+	@ExceptionHandler(ConstraintViolationException.class)
+	public ResponseEntity handle(ConstraintViolationException constraintViolationException) {
```
RichardW98, line 142: ResponseEntity<String>

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/PubliclyVisible.java
```diff
@@ -0,0 +1,21 @@
...
+import javax.validation.Payload;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.TYPE_USE, ElementType.PARAMETER, ElementType.FIELD })
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = PublicVisibilityValidator.class)
+public @interface PubliclyVisible {
```
RichardW98, line 13: we have all other Annotation in parodos-model-api.  should we move this to there?

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java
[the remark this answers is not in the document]
RichardW98: I think we can create globalExceptionhandler some time in the future and move this to there

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkFlowOptionsResponseDTO.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Builder
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class WorkFlowOptionsResponseDTO {
+
+	@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
RichardW98, line 18: move @JsonInclude(JsonInclude.Include.NON_EMPTY) to class level

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
```diff
@@ -194,6 +200,38 @@ public WorkFlowStatusResponseDTO getWorkFlowStatus(UUID workFlowExecutionId) {
...
+		if (params.contains(WorkContextDelegate.Resource.WORKFLOW_OPTIONS)) {
+			options = (Map) WorkContextDelegate.read(workFlowExecution.getWorkFlowExecutionContext().getWorkContext(),
+					WorkContextDelegate.ProcessType.WORKFLOW_EXECUTION, WorkContextDelegate.Resource.WORKFLOW_OPTIONS);
+		}
+		if (options == null) {
```
RichardW98, line 215: init options with `Map.of`

RichardW98: can we define options as WorkFlowOptions as we did before,  and use object mapper to convert 
`WorkContextDelegate.read(workFlowExecution.getWorkFlowExecutionContext().getWorkContext(),
					WorkContextDelegate.ProcessType.WORKFLOW_EXECUTION, WorkContextDelegate.Resource.WORKFLOW_OPTIONS)` to it?

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkFlowContextResponseDTO.java
```diff
@@ -0,0 +1,39 @@
...
+@Builder
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class WorkFlowContextResponseDTO {
+
+	private String workFlowExecutionId;
+
+	@JsonInclude(JsonInclude.Include.NON_EMPTY)
+	private WorkFlowOptionsResponseDTO workFlowOptions;
```
RichardW98, line 37: it looks like this only supports for assessment options,  should we extend it to all publicly visiable fields in workContext?

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/PubliclyVisible.java
masayag: I wouldn't move it if it isn't needed for the user/client.
For now, we use it internally for the purpose of validating input within the service only.
Pls let me know if you see a need for it.

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java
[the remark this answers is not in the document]
masayag: This is a good idea.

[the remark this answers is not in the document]
masayag: Updated https://issues.redhat.com/browse/FLPATH-333 to include this item.

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkFlowContextResponseDTO.java
masayag: Let's do it gradually as needed by the client, else, we'll have to commit to an API that might not be needed, but we'll have to maintain.
We can start with option only, since extending this API doesn't consider breaking it and it will maintain its backward compatibility.

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
masayag: I'm not sure this is the same as before, since when we used to retrieve the `WorkFlowOptions` as part of the `execute()` API, the context remained in memory and wasn't read from the DB.
That means that the actual type of `workFlowExecution.getWorkFlowExecutionContext().getWorkContext(), WorkContextDelegate.ProcessType.WORKFLOW_EXECUTION, WorkContextDelegate.Resource.WORKFLOW_OPTIONS)`  was `WorkflowOptions` as set to it in the workflow-examples.

With this change, the value is being read from the DB, which should convert the json representation of the workflow-execution-context to a WorkFlowOptions.

I don't like the fact we might have more than one type for the same object being held in the context map.
Perhaps this should be solved on the DB layer to convert that part of context from a map to WorkFlowOptions.

Do you have anything specific in mind about how to treat this one?

pkliczewski (commented):
/lgtm

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/PubliclyVisible.java
RichardW98: yes keep it here if it's only to be used for this controller

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/service/WorkFlowServiceImpl.java
RichardW98: my thinking of using WorkFlowOptions is that dto is more structured than Map.  But you are right that from the perspective of whole workcontext,  it's better to keep the value type as `Map`. Maybe in future we can specify the Type for specific Resources(e.g. WORKFLOW_OTPIONS -> WorkFlowOptions).  Let's keep `Map` for now

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkFlowContextResponseDTO.java
RichardW98: ok, let's start with options

RichardW98 (approved):

PR #274, opened 2023-04-24
Refactor Notification-controller API code

nirarg:
This Change is the first step of refactoring and fixing notification-service API
This change include the following:

- Change notification-controller:
Change Get function to return "ResponseEntity<Page<NotificationRecordResponseDTO>>" instead of
    "PagedModel<NotificationRecordResponseDTO>" in order to fix miss matching between the returned value
    to the structure defined in swagger.
Fixes: [FLPATH-260](https://issues.redhat.com/browse/FLPATH-260)
- Change controller unit tests following the changes
- Introduce new notification-service integration test
Currently this include one test validate basic functionality
(related to: [FLPATH-273](https://issues.redhat.com/browse/FLPATH-273))


.github/workflows/test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Maven Test
         id: test
         run: |
-          mvn test --file ./pom.xml
+          mvn test --file ./pom.xml -Dtest=!%regex[.*.integration.*] -DfailIfNoTests=false
```
pkliczewski, line 39: Why do we want to set `failIfNoTests` to false?

Makefile
pkliczewski, line 1: why do we want to set `failIfNoTests` to false in this file?

notification-service-sdk/.openapi-generator/VERSION
```diff
@@ -1 +1 @@
-5.4.0
\ No newline at end of file
+6.2.0
```
pkliczewski, line 1: this change makes it inconsistent with workflow-service-sdk. Please update both

notification-service-sdk/pom.xml
```diff
@@ -174,11 +174,26 @@
...
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <properties>
         <gson-fire-version>1.8.5</gson-fire-version>
         <swagger-core-version>1.6.3</swagger-core-version>
         <jackson-databind-nullable-version>0.2.2</jackson-databind-nullable-version>
-        <openapi.generator.version>5.4.0</openapi.generator.version>
+        <openapi.generator.version>6.2.0</openapi.generator.version>
```
pkliczewski, line 197: please make it consistent with workflow-service-sdk

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
```diff
@@ -59,53 +59,70 @@ public class NotificationRecordController {
...
-		}
-		Page<NotificationRecord> notificationsRecordPage = this.notificationRecordService
-				.getNotificationRecords(pageable, securityUtil.getUsername(), state, searchTerm);
-		return this.notificationRecordPagedResourcesAssembler.toModel(notificationsRecordPage,
-				this.notificationAssemblerDTO);
+
+		Page<NotificationRecord> notificationsPage = this.notificationRecordService.getNotificationRecords(pageable,
+				securityUtil.getUsername(), state, searchTerm);
+		Page<NotificationRecordResponseDTO> notificationsDtoPage = notificationsPage
+				.map(NotificationRecord -> modelMapper.map(NotificationRecord, NotificationRecordResponseDTO.class));
```
eloycoto, line 90: Should we have a method on `NotificationRecordResponseDTO.fromNotificationRecord(NotificationRecord)` we moved to that way in the workflow-service.

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
```diff
@@ -59,53 +59,70 @@ public class NotificationRecordController {
...
+
+		return ResponseEntity.ok(notificationsDtoPage);
 	}
 
 	/**
 	 * Returns unread notifications for a user.
 	 */
+	@io.swagger.v3.oas.annotations.Operation(
+			summary = "Returns the amount of the unread notification records for the user")
```
eloycoto, line 99: Return the number

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
```diff
@@ -59,53 +59,70 @@ public class NotificationRecordController {
...
+	@ApiResponses(value = { @ApiResponse(responseCode = "200", description = "Succeeded"),
+			@ApiResponse(responseCode = "401", description = "Unauthorized"),
+			@ApiResponse(responseCode = "404", description = "Not found") })
+	@io.swagger.v3.oas.annotations.Operation(summary = "Updates the specified notification record with user operation")
 	@PutMapping("/{id}")
 	public NotificationRecordResponseDTO updateNotificationStatusById(@PathVariable("id") UUID id,
 			@RequestParam Operation operation) {
 		NotificationRecord notificationRecord = this.notificationRecordService.updateNotificationStatus(id, operation);
-		return notificationAssemblerDTO.toModel(notificationRecord);
+		return modelMapper.map(notificationRecord, NotificationRecordResponseDTO.class);
```
eloycoto, line 118: TBH, I prefer to use the toModel() is much better for testing.

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
```diff
@@ -0,0 +1,135 @@
...
+		createNotificationMessage("type3", "body3", "subject3", "test2");
+
+		validateNotificationRecordCount(testName, 3);
+		PageNotificationRecordResponseDTO list = validateNotificationRecordList(testName, 3);
+
+		NotificationRecordResponseDTO notificationRecord = list.getContent().get(0);
+
+		logTestStep(testName, "Update one Notification Record as \"READ\"");
+
+		try {
```
eloycoto, line 55: we shouldn't use try in test, should be assertDoesNotThrow

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
```diff
@@ -0,0 +1,135 @@
...
+			log.error("Exception when calling NotificationRecordApi#update:", e);
+			fail("Exception when calling NotificationRecordApi#update");
+		}
+
+		validateNotificationRecordCount(testName, 2);
+		list = validateNotificationRecordList(testName, 3);
+
+		notificationRecord = list.getContent().get(1);
+
+		try {
```
eloycoto, line 68: all the try here shouldn't be here.

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
```diff
@@ -0,0 +1,135 @@
...
+			this.messageApiInstance.create(notificationMessageCreateRequestDTO);
+		}
+		catch (ApiException e) {
+			log.error("Exception when calling NotificationMessageApi#create:", e);
+			fail("Exception when calling NotificationMessageApi#create:");
+		}
+	}
+
+	private void logTestStep(String testName, String stepDescription) {
+		log.info("############## " + testName + ": " + stepDescription);
```
eloycoto, line 132: Is this really needed? if we need to log in a test maybe can be splitted in two?

eloycoto (changes requested):

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
nirarg: I saw that was the way used both for `workflowDefinition` and `project` list (using modelMapper)
If we want to change it, lets address this change for all modules in another PR
(Stop using modelMapper)

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
nirarg: `assertDoesNotThrow` is supported only since junit 5.2, while we use junit 4.13.1
As far as I understand, this can't be updated yet because of Maven supprt
Am I wrong?

eloycoto: we used assertDoesNotThrow in the workflow-service, are u sure about?

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
eloycoto: at the moment you need to add logs inside the test, means that the log may have too many complexity and maybe you can split in more.

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
eloycoto: Why need to be addressed in a different issue?

What I mentioned is to use something like this, that it's much easier to do:
https://github.com/parodos-dev/parodos/blob/main/workflow-service/src/main/java/com/redhat/parodos/workflow/definition/dto/WorkDefinitionResponseDTO.java#L92-L105

nirarg:
/hold
depends on PR #289

notification-service-sdk/.openapi-generator/VERSION
nirarg: Moved this change to different PR
Please see PR #289

notification-service-sdk/pom.xml
nirarg: Moved this change to different PR
Please see PR #289

.github/workflows/test.yml
nirarg: Adding this caused to ignore the default **surefire:test** configuration, as result of that some tests where skipped
Anyway, I changed it to exclude the regex from the pom.xml configuration
Moved this change to different PR, please see PR #290

Makefile
nirarg: See comment above

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
nirarg: Changed
Anyway, need to change adittional places in the code (remove modelMapper usage)

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
nirarg: Fixed

notification-service/src/main/java/com/redhat/parodos/notification/controller/NotificationRecordController.java
nirarg: Changed

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
nirarg: Ok, I was looking for [org.junit.Assert.assertDoesNotThrow](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable))
Imported `org.junit.jupiter` library to `workflow-examples` as well

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
nirarg: Done

nirarg:
The `Build/build` check is failing because need #290 to be merged before

Also #289 commits are included in this PR, please ignore them

Would rebase after both #290 and #289 merges

nirarg:
I changed the integration test file name not to include the `Test` suffix
Will align the test with the changes in PR #281 after it would be merged

workflow-examples/src/test/java/com/redhat/parodos/notification/integration/NotificationRecordTest.java
nirarg: This is integration test, each test may need to validate all operations:
1. Create notification
2. Delete notification
3. get notification
4. count notifications
Therefore, I prefer to have sub function to each operation and than each test function can call to the sub functions, and I created the logs this way to make the flow more clear.
I agree that this is shouldn't be done in unit tests, but this is not the case in this integration test

nirarg:
/hold cancel
The PR is ready for review/merge

rgolangh (commented):
/approve 
the test class could use a custom assertion instead of a function to validate but this is not mandatory

PR #276, opened 2023-04-24
dynamic parameter

RichardW98:
add a controller to dynamically update parameters:
POST.      /api/v1/workflowdefinitions/{workflowDefinitionName}/parameters/update/{valueProviderName}
requestBody:
```
[
    {
        "key": "parameter_key",
        "value": "parameter_value",
        "workName": "onboardingAssessmentTask"
    }
]
```

new field in workParameter: valueProviderName
UI can use this field to decide if this parameter needs dynamic update

example parameter in new workflow definition:
```
"WORKFLOW_SELECT_SAMPLE": {
        "valueProviderName": "complexWorkFlowValueProvider",
        "format": "select",
        "description": "Workflow select parameter sample",
        "type": "string",
        "required": false,
        "enum": [
          "option1",
          "option2"
        ]
      }
```


workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
```diff
@@ -90,4 +100,18 @@ public ResponseEntity<WorkFlowDefinitionResponseDTO> getWorkFlowDefinitionById(@
...
+			value = {
+					@ApiResponse(responseCode = "200", description = "Succeeded",
+							content = { @Content(mediaType = "application/json",
+									schema = @Schema(implementation = List.class)) }),
+					@ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content),
+					@ApiResponse(responseCode = "403", description = "Forbidden", content = @Content) })
+	@PostMapping("/parameters/update/{valueProviderName}")
+	public ResponseEntity<List<WorkParameterValueResponseDTO>> updateParameter(@PathVariable String valueProviderName,
+			@RequestBody List<WorkParameterValueRequestDTO> workParameterValueRequestDTOS) {
+		return ResponseEntity.ok(workFlowParameterService.getValues(valueProviderName, workParameterValueRequestDTOS));
```
eloycoto, line 114: Should we have the parameter a UUID for the workflow definition ID? So if the parameter is the same for the same WD is just weird

workflow-service/src/test/java/com/redhat/parodos/workflow/definition/parameter/WorkFlowParameterServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@ExtendWith(SpringExtension.class)
+class WorkFlowParameterServiceImplTest {
+
+	private static final String DATA_PROVIDER_NAME = "test-data-provider";
+
+	private static final String INVALID_DATA_PROVIDER = "invalid_data_provider";
```
anludke, line 26: use "-" for consistency

parodos-model-api/src/main/java/com/redhat/parodos/workflow/parameter/WorkFlowValueProvider.java
```diff
@@ -0,0 +1,9 @@
+package com.redhat.parodos.workflow.parameter;
+
+import java.util.List;
+
+public interface WorkFlowValueProvider {
```
anludke, line 5: Shouldn't it be _WorkFlowParameterValueProvider_? But also if we'd like to keep it generic for workflow and workflow task parameter, it might be better to name it as _WorkParameterValueProvider_

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
RichardW98: yes,  it should an existing parameter from WD.   users' data providers can use the parameter-key to get the parameter data from db.  key should be unique for parameters within the same work.  UUID can also work.

anludke (changes requested):
Please add tests for the new endpoint :)

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
```diff
@@ -90,4 +100,18 @@ public ResponseEntity<WorkFlowDefinitionResponseDTO> getWorkFlowDefinitionById(@
...
 
+	@Operation(summary = "Returns updated parameter value")
+	@ApiResponses(
+			value = {
+					@ApiResponse(responseCode = "200", description = "Succeeded",
+							content = { @Content(mediaType = "application/json",
+									schema = @Schema(implementation = List.class)) }),
+					@ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content),
+					@ApiResponse(responseCode = "403", description = "Forbidden", content = @Content) })
+	@PostMapping("/parameters/update/{valueProviderName}")
```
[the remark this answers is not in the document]
RichardW98: the request and response can contain multiple parameters

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
RichardW98: @eloycoto  Maybe we can also create a  **Parameter** table  and map it to workflowDefinition & workflowTaskDefinition table. so that records in Parameters can also have relations e.g. allof, anyof ....  And of course UUID will make more sense in this way

workflow-service-sdk/api/openapi.yaml
```diff
@@ -127,6 +127,41 @@ paths:
       tags:
       - Workflow Definition
       x-accepts: application/json
+  /api/v1/workflowdefinitions/parameters/update/{valueProviderName}:
```
eloycoto, line 130: I'm worried here. What happens if a workflow has the list of users retrieved from the database? Are we exposing those parameters to all of the workflows? Shouldn't we have a mechanism that does not allow it?

I'm a bit worried about this, because users maybe trust in the thing, and it's insecure by default.

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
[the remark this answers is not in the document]
eloycoto: WorkFlowDefinitionID/parameters/update/{valueProviderName}

eloycoto (changes requested):

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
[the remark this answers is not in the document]
RichardW98: so far the workName is unique throughout the whole application,   I think the workName in **WorkParameterValueRequestDTO** is good enough to detect which workflow it is requesting to.

workflow-service-sdk/api/openapi.yaml
RichardW98: https://github.com/parodos-dev/parodos/pull/276/files#diff-087771281f17597fca047d56f3fd3e6e3ca16146e578817ca6ad2aa850245e13R18 the request to this controller can specify workName.  ValueProvider can use it to validate if the parameter is from a validated work in the workflow

eloycoto: I'm not sure how that can fix a possible security issue :\

RichardW98: ok, I will change the controller to WorkFlowDefinitionID/parameters/update/{valueProviderName} and use the workflowDefinition to validate the workflow requested.

eloycoto: happy to track that on the jira, but keep the idea in your mind

RichardW98: the controller has been updated to '/{workflowDefinitionName}/parameters/update/{valueProviderName}'

pkliczewski:
@RichardW98 for some reason I thought we wanted to fetch param values from third party system. Is it something we want to do it in the future?

RichardW98:
> @RichardW98 for some reason I thought we wanted to fetch param values from third party system. Is it something we want to do it in the future?

yes,  this pr is meant to fetch the values from 3rd-party system with the ValueProvider   based on UI input

workflow-service-sdk/api/openapi.yaml
RichardW98: the jira ticket is created: https://issues.redhat.com/browse/FLPATH-336

pkliczewski:
lgtm

anludke (approved):

anludke (commented):
/lgtm

parodos-model-api/src/main/java/com/redhat/parodos/workflow/parameter/WorkParameter.java
```diff
@@ -48,16 +50,24 @@ public class WorkParameter {
...
 			return Map.of();
 		}
-		Map<String, Object> properties = this.getType().getAsJsonSchema();
-		properties.put("required", !this.isOptional());
-		properties.put("description", this.getDescription());
+		Map<String, Object> properties = type.getAsJsonSchema();
+		properties.put("required", !optional);
+		properties.put("description", description);
+		if (valueProviderName != null && !valueProviderName.isEmpty())
+			properties.put("valueProviderName", valueProviderName);
```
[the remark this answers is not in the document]
RichardW98: if in the future we turn to persist it in db, then UUID will work here

PR #283, opened 2023-04-26
NO-ISSUE: Make target to automate releases

pkliczewski:
New target `release-all` automates release process as well as builds, tags and pushes new version of images.

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [x] CI
- [ ] Documentation
- [ ] Auto generated SDK code

**Impacted services**
- [ ] Workflow Service
- [ ] Notifivcation Service

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ NOTIFICATION_SERVICE_IMAGE=notification-service
 
 # get version from pom
 VERSION = $(shell sed -n "s/<revision>\(.*\)<\/revision>/\1/p" $(PWD)/pom.xml | sed 's/\ *//g')
+RELEASE_VERSION = $(shell echo $(VERSION) | sed -e s/-SNAPSHOT//)
+NEXT_VERSION = $(shell echo $(RELEASE_VERSION) | awk -F. -v OFS=. '{$$NF += 1 ; print}')-SNAPSHOT
 
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed s,^main$$,latest,g)
 GIT_HASH := $(shell git rev-parse HEAD)
@@ -74,7 +76,10 @@ FAST_BUILD_ARGS = -Dmaven.test.skip=true -Dmaven.javadoc.skip=true
 clean: ## Clean all modules
 	$(MAVEN) clean
 
-all: mvn-checks java-checks clean ## Build all modules
+all: mvn-checks java-checks install ## Build all modules
+
+install: ARGS = $(FAST_BUILD_ARGS)
+install: clean
 	$(MAVEN) $(ARGS) install
 
 deploy: clean ## push snapshot modules to maven central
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -130,23 +135,25 @@ fast-build-coverage: coverage
...
 tag-images: ## Tag docker images with git hash and branch name
-	$(DOCKER) tag docker-compose_workflow-service:latest $(ORG)$(WORKFLOW_SERVICE_IMAGE):$(GIT_HASH)
-	$(DOCKER) tag docker-compose_notification-service:latest $(ORG)$(NOTIFICATION_SERVICE_IMAGE):$(GIT_HASH)
+	$(eval TAG?=$(GIT_HASH))
+	$(DOCKER) tag docker-compose_workflow-service:latest $(ORG)$(WORKFLOW_SERVICE_IMAGE):$(TAG)
+	$(DOCKER) tag docker-compose_notification-service:latest $(ORG)$(NOTIFICATION_SERVICE_IMAGE):$(TAG)
 
-	$(DOCKER) tag docker-compose_workflow-service:latest $(ORG)$(WORKFLOW_SERVICE_IMAGE):$(GIT_BRANCH)
-	$(DOCKER) tag docker-compose_notification-service:latest $(ORG)$(NOTIFICATION_SERVICE_IMAGE):$(GIT_BRANCH)
+	$(DOCKER) tag docker-compose_workflow-service:latest $(ORG)$(WORKFLOW_SERVICE_IMAGE):$(TAG)
```
eloycoto, line 147: If this TAG we need to change in `.github/workflows/`

Makefile
```diff
@@ -166,6 +173,27 @@ wait-kubernetes-dependencies: ## Wait for dependencies to be ready
 	  --selector=app.kubernetes.io/component=controller \
 	  --timeout=90s
 
+update-version: ## update release version
+	find . -type f | xargs sed -i 's/$(VERSION)/${RELEASE_VERSION}/g'
+
+bump-version: ## update post-release version and update commit message
+	find . -type f | xargs sed -i 's/$(RELEASE_VERSION)/${NEXT_VERSION}/g'
+
+git-commit: ## adds all files and bumps the version
```
nirarg, line 182: This is too general target name, need to have name that clarify it only for version bump
suggestion: `bump-git-commit`
Or maybe this target can be merged with `bump-version` target?

Makefile
```diff
@@ -166,6 +173,27 @@ wait-kubernetes-dependencies: ## Wait for dependencies to be ready
...
+	git add .
+	git commit -m 'Version bump'
```
masayag, line 184: Can we include also the version we bump to? (should ease the search when/if needed).

Makefile
```diff
@@ -74,7 +76,10 @@ FAST_BUILD_ARGS = -Dmaven.test.skip=true -Dmaven.javadoc.skip=true
 clean: ## Clean all modules
 	$(MAVEN) clean
 
-all: mvn-checks java-checks clean ## Build all modules
+all: mvn-checks java-checks install ## Build all modules
+
+install: ARGS = $(FAST_BUILD_ARGS)
+install: clean
```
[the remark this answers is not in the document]
pkliczewski: `mvn install` influenced the name of this target to be `install`. Running all in release sequence fails.

Makefile
pkliczewski: The changes do not influence current behavior. Please see

> $(eval TAG?=$(GIT_HASH))

We still use hash and we only use TAG when we run `release-all` target

Makefile
pkliczewski: Will change the name. Merging both targets is not good idea because they do not run one after the other.

Makefile
pkliczewski: Sure we can. I based the flow on current way of releasing. It makes sense to have version there.

Makefile
```diff
@@ -166,6 +173,27 @@ wait-kubernetes-dependencies: ## Wait for dependencies to be ready
...
 	  --timeout=90s
 
+update-version: ## update release version
+	find . -type f | xargs sed -i 's/$(VERSION)/${RELEASE_VERSION}/g'
+
+bump-version: ## update post-release version and update commit message
+	find . -type f | xargs sed -i 's/$(RELEASE_VERSION)/${NEXT_VERSION}/g'
+
+git-commit: ## adds all files and bumps the version
+	git add .
```
gciavarrini, line 183: May be a safer choice, use `git add -u .`? So it will commit only tracked files

pkliczewski: Good point, at this time we do not expect any files which are untracked. Will fix.

Makefile
gciavarrini: Overall lgtm, only one question: is it ok that the commit is unsigned?

Makefile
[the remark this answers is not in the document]
nirarg: So one of the `mvn-checks` or `java-checks` fails in the release sequence? why?

Makefile
```diff
@@ -18,6 +18,8 @@ NOTIFICATION_SERVICE_IMAGE=notification-service
 
 # get version from pom
 VERSION = $(shell sed -n "s/<revision>\(.*\)<\/revision>/\1/p" $(PWD)/pom.xml | sed 's/\ *//g')
```
RichardW98, line 20: I think it's safer to use something like 'mvn help:evaluate -Dexpression=project.version -q -DforceStdout' to get the project version. but if we can't use mvn cmd here then this is fine

pkliczewski: I will explore

pkliczewski: I could use mvn here but we would not be able to run `make mvn-checks` if maven not installed. Is there any other non maven command we could use here?

RichardW98: I don't think so. let's keep sed then

PR #284, opened 2023-04-26
fix: backstage app-config update

eloycoto:
This will fix the backstage deployment and be able to deploy backstage.

### DIFF
--- a/hack/manifests/backstage/app-config.yaml
+++ b/hack/manifests/backstage/app-config.yaml
@@ -1,6 +1,10 @@
 app:
   title: Scaffolded Backstage App
   baseUrl: http://0.0.0.0:3000
+parodos:
+  workflows:
+    assessment: 'onboardingAssessment_ASSESSMENT_WORKFLOW'
+    assessmentTask: 'onboardingAssessmentTask'
 
 organization:
   name: My Company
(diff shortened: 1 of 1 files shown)


hack/manifests/backstage/app-config.yaml
```diff
@@ -64,6 +68,16 @@ proxy:
       accept: 'application/json'
       Authorization: 'Basic dGVzdDp0ZXN0'
 
+  '/parodos-notifications':
+    target: 'http://notification-service.default.svc.cluster.local:8080/api/v1'
```
masayag, line 72: isn't the port for notification-service 8081?

hack/manifests/backstage/app-config.yaml
```diff
@@ -1,6 +1,10 @@
 app:
   title: Scaffolded Backstage App
   baseUrl: http://0.0.0.0:3000
+parodos:
```
masayag, line 4: what is this configuration responsible for?

RichardW98: it is to define the workflow and its tasks for assessment. Currently it has to be configured in this way. in the future we will have a controller to fetch it from backend

PR #289, opened 2023-04-30
Services SDK: Change openapi-generator version to 6.4.0

nirarg:
This change is done in both "workflow-service-sdk" and "notification-service-sdk"
    Include:
    * Move version properties to parent pom.xml, to have one version for both modules
    * Ignore UnitTests generation in sdk modules, this is done to avoid error after openapi-generator version change
      (the unit tests in those modules are empty and not used)

Related to PR #274

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [x] Auto generated SDK code

**Impacted services**
- [x] Workflow Service
- [x] Notifivcation Service

**Checklist**


rgolangh:
not sure about the pom additions - can you double check if its all needed by the bump of the version?

nirarg:
> not sure about the pom additions - can you double check if its all needed by the bump of the version?

Removed all unused dependencies

workflow-service-sdk/src/main/java/com/redhat/parodos/sdk/model/ArgumentRequestDTO.java
```diff
@@ -19,16 +19,32 @@
...
+import java.util.Set;
+
+import com.redhat.parodos.sdk.invoker.JSON;
+
 /**
  * ArgumentRequestDTO
  */
-@lombok.Data
-@lombok.AllArgsConstructor
-@lombok.Builder
```
masayag, line 31: Is it intentional? Removing all lombok annotations?

nirarg: This is auto generated code, it was removed after changing the generator version

masayag: I see it was removed explicitly from here:

https://github.com/parodos-dev/parodos/pull/289/files#diff-2fdbd8c157a3ebaa9bccc9812c554feeb04a5cd3b70f8b3222f63cdb3ad6d4dcL156-L159

Wouldn't it work if this one is restrored?

masayag: To be exact - this is the change in the existing code base which is removed in this PR:
https://github.com/parodos-dev/parodos/blob/main/workflow-service-sdk/pom.xml#L156

gciavarrini: Lombok's annotations were explicitly added to have builder, getters, setters, and constructor in the generated classes. This eases the interaction with the SDK classes. Can we keep them?

workflow-service-sdk/.openapi-generator-ignore
```diff
@@ -21,4 +21,6 @@
 #docs/*.md
 # Then explicitly reverse the ignore rule for a single file:
 #!docs/README.md
-pom.xml
\ No newline at end of file
+pom.xml
+src/test/**/*.java
```
gciavarrini, line 25: I think we should skip the generation from plugin configuration adding
```
<generateApiTests>false</generateApiTests>
<generateModelTests>false</generateModelTests>
```

workflow-service-sdk/README.md
```diff
@@ -12,7 +12,7 @@ This is the API documentation for the Parodos Workflow Service. It provides oper
 ## Requirements
 
 Building the API client library requires:
-1. Java 1.7+
+1. Java 1.8+
```
gciavarrini, line 15: Aren't we using Java 11 ?

workflow-service-sdk/pom.xml
```diff
@@ -171,62 +166,22 @@
...
             <artifactId>validation-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-web</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.projectlombok</groupId>
-            <artifactId>lombok</artifactId>
-            <scope>provided</scope>
```
gciavarrini, line 229: As I said, I don't see the value of removing Lombok's annotations

workflow-service-sdk/pom.xml
```diff
@@ -171,62 +166,22 @@
...
+            <groupId>jakarta.ws.rs</groupId>
+            <artifactId>jakarta.ws.rs-api</artifactId>
```
gciavarrini, line 184: This plugin replaces the deleted plugins? Is it suggested from open api generator?

workflow-service-sdk/README.md
nirarg: We are using Java 17
This is auto generated README for the SDK requirements

gciavarrini: You're right! Thanks for the clarification!

PR #295, opened 2023-05-02
Add constraints on model entities names

masayag:
Users and projects should be uniquely identified by their names in the system.
To enforce this, unique constraints should be implemented at the database level.
When requests are made to the controller's endpoints, they should fail with an HTTP status code of 409 (Conflict) if the constraints are violated.

To simplify JPA testing, `RepositoryTestBase` was added to serve as a base test class for the *Repository test classes.

**Which issue(s) this PR fixes** *:
Fixes: #[FLPATH-341](https://issues.redhat.com/browse/FLPATH-341)

**Change type**
- [ ] New feature
- [x] Bug fix
- [x] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto generated SDK code

**Impacted services**
- [x] Workflow Service
- [ ] Notifivcation Service

**Checklist**
- [x] Relevant issues have been referenced.

### DIFF
--- a/workflow-service/src/main/java/com/redhat/parodos/project/controller/ProjectController.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/project/controller/ProjectController.java
@@ -15,14 +15,14 @@
  */
 package com.redhat.parodos.project.controller;
 
-import com.redhat.parodos.project.dto.ProjectRequestDTO;
-import com.redhat.parodos.project.dto.ProjectResponseDTO;
-import com.redhat.parodos.project.service.ProjectServiceImpl;
 import javax.validation.Valid;
 import java.net.URI;
 import java.util.List;
 import java.util.UUID;
 
+import com.redhat.parodos.project.dto.ProjectRequestDTO;
+import com.redhat.parodos.project.dto.ProjectResponseDTO;
+import com.redhat.parodos.project.service.ProjectServiceImpl;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Content;
--- a/workflow-service/src/main/java/com/redhat/parodos/project/service/ProjectServiceImpl.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/project/service/ProjectServiceImpl.java
@@ -15,8 +15,10 @@
  */
 package com.redhat.parodos.project.service;
 
+import javax.persistence.EntityExistsException;
 import java.util.Date;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -54,6 +56,11 @@ public class ProjectServiceImpl implements ProjectService {
 
 	@Override
 	public ProjectResponseDTO save(ProjectRequestDTO projectRequestDTO) {
+		Optional<Project> projectByName = projectRepository.findByNameIgnoreCase(projectRequestDTO.getName());
+		if (projectByName.isPresent()) {
+			throw new EntityExistsException(
+					String.format("Project with name: %s already exists", projectByName.get().getName()));
+		}
 		// get user from security utils and set on project
 		Project project = projectRepository.save(Project.builder().name(projectRequestDTO.getName())
 				.description(projectRequestDTO.getDescription()).createDate(new Date()).modifyDate(new Date()).build());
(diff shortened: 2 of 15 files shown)


workflow-service/src/main/java/com/redhat/parodos/project/controller/ProjectController.java
```diff
@@ -63,7 +65,8 @@ public ProjectController(ProjectServiceImpl projectService) {
 			@ApiResponse(responseCode = "201", description = "Created",
 					content = { @Content(mediaType = "application/json",
 							schema = @Schema(implementation = ProjectResponseDTO.class)) }),
-			@ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content) })
+			@ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content),
+			@ApiResponse(responseCode = "409", description = "Conflict", content = @Content) })
```
eloycoto, line 69: When it should return 409? Should we add a test for this?

masayag: I'll address the exception handling from the rest-controller as part of introducing a global exception handler as part of https://issues.redhat.com/browse/FLPATH-342.
I didn't want to implement this within this patch, in order not to overload it with changes.
I don't mind excluding it from this PR.

masayag: I removed the API changes and will address them in the next PRs

workflow-service/src/main/java/com/redhat/parodos/project/service/ProjectServiceImpl.java
```diff
@@ -54,6 +55,10 @@ public ProjectServiceImpl(ProjectRepository projectRepository, WorkFlowRepositor
 
 	@Override
 	public ProjectResponseDTO save(ProjectRequestDTO projectRequestDTO) {
+		if (projectRepository.findByName(projectRequestDTO.getName()).isPresent()) {
+			throw new EntityExistsException(
+					String.format("Project with name: %s already exists", projectRequestDTO.getName()));
+		}
 		// get user from security utils and set on project
 		Project project = projectRepository.save(Project.builder().name(projectRequestDTO.getName())
 				.description(projectRequestDTO.getDescription()).createDate(new Date()).modifyDate(new Date()).build());
```
anludke, line 64: Moving forwards, what do you think of setting createDate and modifyDate via sysdate at the db level?

workflow-service/src/main/java/com/redhat/parodos/project/service/ProjectServiceImpl.java
```diff
@@ -54,6 +55,10 @@ public ProjectServiceImpl(ProjectRepository projectRepository, WorkFlowRepositor
 
 	@Override
 	public ProjectResponseDTO save(ProjectRequestDTO projectRequestDTO) {
+		if (projectRepository.findByName(projectRequestDTO.getName()).isPresent()) {
```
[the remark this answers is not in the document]
masayag: Do you think we should enforce it on DB level as well?

workflow-service/src/main/java/com/redhat/parodos/project/service/ProjectServiceImpl.java
masayag: That's a tricky one :-) 
It depends if the clock of the Postgres-db server is in sync with the workflow service.
If the importance is when the request is sent by the user - then I'd keep the current code.
If the importance is the actual record creation time - then use sysdate.

workflow-service/src/main/java/com/redhat/parodos/user/repository/UserRepository.java
```diff
@@ -28,6 +29,6 @@
 
 public interface UserRepository extends JpaRepository<User, UUID> {
 
-	List<User> findByUsername(String username);
+	Optional<User> findByUsername(String username);
```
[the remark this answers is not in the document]
masayag: How users are being created in the system? will there be a need to import users from other systems?

workflow-service/src/main/java/com/redhat/parodos/project/service/ProjectServiceImpl.java
anludke: Agreed with your points! Since we don't specific requirements on this, let's keep as-is :)
We may revisit later

anludke (approved):

anludke (commented):
/lgtm

PR #296, opened 2023-05-02
simpler status usage

rgolangh:
- Fixing my mistake from #293 
- Using now the enum value in all places 
- Added an interceptor test to start the coverage of those areas, specifically this change. (more needed)


masayag:
I was under the impression that we shouldn't modify code in the workflow-engine module as it is imported from easy-flow project. Can we change it?

rgolangh:
easy-flow is in readonly mode, not being further developed. https://github.com/parodos-dev/parodos/blob/main/workflow-engine/README.md#comments-on-easy-flow-usage

PR #329, opened 2023-05-09
Fail correctly when execution fails

masayag:
When an execution of a workflow fails due to incorrect input or due to internal error, a report with the error should be returned to the user. The status of the report in that case will be FAILED and the HTTP request code will be 500, for now.

**Change type**
- [ ] New feature
- [x] Bug fix
- [x] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [x] Workflow Service
- [ ] Notification Service

**Checklist**


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/controller/WorkFlowController.java
```diff
@@ -84,7 +85,7 @@ public WorkFlowController(WorkFlowService workFlowService) {
 	@PostMapping
 	public ResponseEntity<WorkFlowResponseDTO> execute(@RequestBody @Valid WorkFlowRequestDTO workFlowRequestDTO) {
 		WorkReport workReport = workFlowService.execute(workFlowRequestDTO);
-		if (workReport == null) {
+		if (workReport.getStatus() == WorkStatus.FAILED) {
```
gciavarrini, line 88: To be sure: `workReport` can't be null?

PR #338, opened 2023-05-16
Use global exception handler by REST controllers

masayag:

In order to activate the validation in the rest-controllers, a dedicated annotation @Validated needs to be added.

By doing so and with early work to enable the validator, the already @Valid annotations, and others will be considered when rest calls are being submitted.

In addition, a global exception handler is added, for now with the functionality of capturing any ResourceNotFound sort of exception.

Gradually, we'll remove controller-aware context from the service layer, so StatusResponse... will be defined in the context of the controllers and not within the services.

Fixes #[FLPATH-333](https://issues.redhat.com/browse/FLPATH-333)
[FLPATH-342](https://issues.redhat.com/browse/FLPATH-342)

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [ ] Workflow Service
- [ ] Notification Service

**Checklist**


workflow-service/src/main/java/com/redhat/parodos/project/controller/ProjectController.java
```diff
@@ -53,6 +54,7 @@
 
 @CrossOrigin(origins = "*", maxAge = 1800)
 @RestController
+@Validated
```
gciavarrini, line 57: Do you think the test should be updated?

workflow-service/src/main/java/com/redhat/parodos/security/LoginController.java
```diff
@@ -35,6 +36,7 @@
 
 @CrossOrigin(origins = "*", maxAge = 1800)
 @RestController
+@Validated
```
gciavarrini, line 39: Do you think it's worth to add a test for this class?

workflow-service/src/main/java/com/redhat/parodos/workflow/definition/controller/WorkFlowDefinitionController.java
```diff
@@ -53,6 +54,7 @@
 
 @CrossOrigin(origins = "*", maxAge = 1800)
 @RestController
+@Validated
```
gciavarrini, line 57: Do you think the test should be updated?

pkliczewski (commented):
I think Gloria raised good point. Would it make sense to update existing tests?

workflow-service/src/main/java/com/redhat/parodos/project/controller/ProjectController.java
masayag: yes, but pls let me add them as integration tests, after your PR is merged.

masayag:
> I think Gloria raised good point. Would it make sense to update existing tests?

yes, but I'd like to add them as integration tests since this is the controller and I'd like to test the actual API instead of mocking it.

pkliczewski (commented):
/lgtm

workflow-service/src/main/java/com/redhat/parodos/common/controller/ControllerExceptionHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+import org.springframework.web.context.request.WebRequest;
+
+@RestControllerAdvice
+public class ControllerExceptionHandler {
```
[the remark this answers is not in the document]
masayag: I wouldn't like to reuse the same code between the two service. It is too coupled to the workflow service and as it will expand, more will be added to it which has relevancy to the workflow service itself.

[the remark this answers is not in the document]
pkliczewski: make sense, thank you. Let's think about it later

PR #341, opened 2023-05-16
Add a workflow task that uses REST

ydayagi:
[FLPATH-144](https://issues.redhat.com//browse/FLPATH-144) https://issues.redhat.com/browse/FLPATH-214
The task can be used as a base class for other/enhanced REST tasks. The protected methods enable that. These methods enable custom building of request headers and custom response processing.

### DIFF
--- a/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
+++ b/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
@@ -1,0 +1,127 @@
+package com.redhat.parodos.tasks.rest;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import com.redhat.parodos.workflow.exception.MissingParameterException;
+import com.redhat.parodos.workflow.parameter.WorkParameter;
+import com.redhat.parodos.workflow.parameter.WorkParameterType;
+import com.redhat.parodos.workflow.task.BaseWorkFlowTask;
+import com.redhat.parodos.workflow.task.enums.WorkFlowTaskOutput;
+import com.redhat.parodos.workflows.work.DefaultWorkReport;
+import com.redhat.parodos.workflows.work.WorkContext;
+import com.redhat.parodos.workflows.work.WorkReport;
+import com.redhat.parodos.workflows.work.WorkStatus;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.RestClientException;
+
+@Slf4j
+public class RestWorkFlowTask extends BaseWorkFlowTask {
+
+	private RestService restService;
+
+	public RestWorkFlowTask() {
+		restService = new RestServiceImpl();
+	}
+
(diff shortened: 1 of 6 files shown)


prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	RestWorkFlowTask(String beanName, RestService restService) {
+		this.restService = restService;
+		setBeanName(beanName);
+	}
+
+	@Override
+	public @NonNull List<WorkParameter> getWorkFlowTaskParameters() {
+		LinkedList<WorkParameter> params = new LinkedList<>();
```
pkliczewski, line 41: please add TLS support

pkliczewski (commented):
The change looks good. Please add tls support

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
```diff
@@ -0,0 +1,127 @@
...
+
+	protected HttpEntity<String> buildRequestEntity(WorkContext workContext) throws RestClientException {
+		String content = getOptionalParameterValue(workContext, "content", "");
+		return new HttpEntity<>(content, buildHttpHeaders(workContext));
+	}
+
+	protected HttpHeaders buildHttpHeaders(WorkContext workContext) throws RestClientException {
+		HttpHeaders httpHeaders = new HttpHeaders();
+
+		httpHeaders.setContentType(MediaType.APPLICATION_JSON);
```
masayag, line 109: should it be configurable?

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
```diff
@@ -0,0 +1,127 @@
...
+	RestWorkFlowTask(String beanName, RestService restService) {
+		this.restService = restService;
+		setBeanName(beanName);
+	}
+
+	@Override
+	public @NonNull List<WorkParameter> getWorkFlowTaskParameters() {
+		LinkedList<WorkParameter> params = new LinkedList<>();
+		params.add(WorkParameter.builder().key("url").type(WorkParameterType.TEXT).optional(false)
+				.description("URL to send request to").build());
```
masayag, line 43: not sure how it is in other places, but I'd extract the keys to constants or to a parameter class.
the need to repeat the exact key name and be aware of its type isn't great.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
ydayagi: we provide basic/general/common config as a base class. user can use it as is or extend.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/rest/RestWorkFlowTask.java
ydayagi: I do not understand what u mean by 'repeat the key' and 'aware of its type'.
The key is defined with its type only once.
The user has to be aware of the type when the request is created.

masayag:
@ydayagi would you like to add integration tests as part of this PR or in a new one and let this be merged?

ydayagi:
> @ydayagi would you like to add integration tests as part of this PR or in a new one and let this be merged?

merge

PR #370, opened 2023-05-24
Fix: return workreport from running the escalation workflow

jordigilh:
I noticed that when running a check flow that has an escalation flow associated to it, the parent `executer` does not return the workreport from running the escalation flow. This PR fixes that so that the return workreport is from the escalation flow and not from the check flow.

@lshannon @RichardW98 PTAL.

/Jordi

### DIFF
--- a/parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/checker/BaseWorkFlowCheckerTask.java
+++ b/parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/checker/BaseWorkFlowCheckerTask.java
@@ -75,7 +75,7 @@ public abstract class BaseWorkFlowCheckerTask extends BaseWorkFlowTask {
 		if (escalationWorkflow != null && report.getStatus() == WorkStatus.FAILED
 				&& new Date().getTime() > expectedCompletionDate) {
 			// run escalation if SLA is exceeded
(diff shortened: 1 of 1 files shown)


parodos-model-api/src/main/java/com/redhat/parodos/workflow/task/checker/BaseWorkFlowCheckerTask.java
```diff
@@ -75,7 +75,7 @@ public WorkReport execute(WorkContext workContext) {
 		if (escalationWorkflow != null && report.getStatus() == WorkStatus.FAILED
 				&& new Date().getTime() > expectedCompletionDate) {
 			// run escalation if SLA is exceeded
-			WorkFlowEngineBuilder.aNewWorkFlowEngine().build().run(escalationWorkflow, workContext);
+			return WorkFlowEngineBuilder.aNewWorkFlowEngine().build().run(escalationWorkflow, workContext);
```
RichardW98, line 78: this is the engine to run escalation.  the returned report is not checker's report

jordigilh: ah, so where is the return value captured from the escalation run?

jordigilh:
As discussed, it's working as expected so I'm closing this PR :smile:

PR #383, opened 2023-05-29
feat: move backstage to sqlite

eloycoto:
Other day Paul mentioned that you can use sqlite, to avoid the issues with  postgres we can use the following configuration to avoid demo errors


eloycoto:
@gciavarrini  wdyt?

pkliczewski:
Would it be possible to keep both? I think we can learn about how to use postgresql well [here](https://backstage.io/docs/deployment/k8s#creating-the-postgresql-database)

gciavarrini:
> Would it be possible to keep both? I think we can learn about how to use postgresql well [here](https://backstage.io/docs/deployment/k8s#creating-the-postgresql-database)

@pkliczewski postgress seems to workfine after this changes https://github.com/gciavarrini/parodos/tree/fix-openshift-postegres 
Not sure if this can be helpful for this PR

pkliczewski:
> @pkliczewski postgress seems to workfine after this changes https://github.com/gciavarrini/parodos/tree/fix-openshift-postegres Not sure if this can be helpful for this PR

I would love to keep both

eloycoto:
The thing is that with the changes that @gciavarrini introduces may not work on kind, I would like to be as simple as possible.  Having postgres in there only adds some burden and maybe a user will use like that

eloycoto:
> @eloycoto what is our recommended way to deploy backstage? do we want customers to use sqlite?

Do we want our customers to use a single pod with a postgres without backup/failover/replication etc..? The Postgres that we have it's just for running backstage, we no longer need that.. so I would drop it. 

Having two backstage means that we need to keep two files up to date, moving things around, plus all the problems that @gciavarrini discovered on running an example Postgres in openshift.

pkliczewski:
@eloycoto do we still need this change?

eloycoto:
I would love to have that :D let's keep this on-hold.

PR #376, opened 2023-05-26
Enable notification integrationtest

gciavarrini:
**What this PR does / why we need it**:
Enable notification service integration tests.
`NotificationRecordTest` has been refactored to match project code style/implementation.
In addition to that, it was necessary to fix the test to let it work with the integration tests in CI.

Future PRs will improve the `NotificationRecordTest` implementation.

Documentation PR available soon.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story (FLPATH-xxxx)*:
Fixes # [FLPATH-348](https://issues.redhat.com/browse/FLPATH-348)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Unit tests
- [x] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [x] Workflow Service
- [x] Notification Service

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.

PR #391, opened 2023-05-31
A notifier bean for interfacing with the notification service

rgolangh:
Tasks of workflows need a builtin conveient way of sending notifications
and for a lot of reasons it makes more sense then a special notificaion
tasks - specially when the input for the message exists in the task
context and its waste or just too much to pass on the context for a
special notification task. It also doesn't need plumbing the task into a
workflow which is combersome.

To send a notificaion just inject a Notifier into a bean:

@Bean
SomeTask createSomeTask(Notifier notifier) {
      new SomeTask(notifier)...
}

class SomeTask {
    SomeTask(Notifier notifier) { this.notifier = notifier; ... }

    execute(WorkContext ctx) {
        ... //work
        var m = new NotificationMessageCreateRequestDTO();
        m.setBody("body");
        notifier.send(m);
    }
}

### DIFF
--- a/hack/manifests/testing/parodos-patch.yaml
+++ b/hack/manifests/testing/parodos-patch.yaml
@@ -19,7 +19,7 @@ data:
   DATASOURCE_USERNAME: "parodos"
   DATASOURCE_PASSWORD: "parodos"
   SPRING_PROFILES_ACTIVE: "dev"
-  NOTIFICATION_SERVER_ADDRESS: "notification-service.default.svc.cluster.local"
+  NOTIFICATION_SERVER_URL: "http://notification-service.default.svc.cluster.local:8080"
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/workflow-service/src/main/java/com/redhat/parodos/NotificationSender.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/NotificationSender.java
@@ -1,0 +1,47 @@
+package com.redhat.parodos;
+
+import java.util.Base64;
+
+import javax.ws.rs.core.HttpHeaders;
+
+import com.redhat.parodos.notification.sdk.api.ApiClient;
+import com.redhat.parodos.notification.sdk.api.ApiException;
+import com.redhat.parodos.notification.sdk.api.NotificationMessageApi;
+import com.redhat.parodos.notification.sdk.model.NotificationMessageCreateRequestDTO;
+import com.redhat.parodos.workflow.task.infrastructure.Notifier;
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+public class NotificationSender implements Notifier {
+
+	private final NotificationMessageApi client;
+
+	public NotificationSender(@Value("${notification.url}") String url,
+			@Value("${notification.auth.basic.user}") String user,
+			@Value("${notification.auth.basic.password}") String password) {
+		ApiClient apiClient = new com.redhat.parodos.notification.sdk.api.ApiClient().setBasePath(url).addDefaultHeader(
+				HttpHeaders.AUTHORIZATION,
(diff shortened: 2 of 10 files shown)


workflow-service/src/main/resources/application-local.yml
```diff
@@ -33,3 +33,13 @@ logging:
...
 
+notification:
+  url: "${NOTIFICATION_SERVICE_URL:http://localhost:8081}"
+  auth:
+    basic:
+      user: test
+      password: test
+
+workflows:
+  mta:
```
eloycoto, line 44: not sure if this (MATA) should be here, or should be part of workflow-examples/src/main/resources/

workflow-service/src/main/java/com/redhat/parodos/NotificationSender.java
```diff
@@ -0,0 +1,40 @@
...
+import com.redhat.parodos.notification.sdk.api.ApiException;
+import com.redhat.parodos.notification.sdk.api.NotificationMessageApi;
+import com.redhat.parodos.notification.sdk.model.NotificationMessageCreateRequestDTO;
+import com.redhat.parodos.workflow.task.infrastructure.Notifier;
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.beans.factory.annotation.Value;
+
+@Slf4j
+@org.springframework.context.annotation.Configuration
```
RichardW98, line 17: maybe @Component is better

workflow-service/src/main/java/com/redhat/parodos/NotificationSender.java
```diff
@@ -0,0 +1,46 @@
...
+				"Basic " + Base64.getEncoder().encodeToString((user + ":" + password).getBytes()));
+		client = new NotificationMessageApi(apiClient);
+	}
+
+	public void send(NotificationMessageCreateRequestDTO message) {
+		try {
+			trySend(message);
+		}
+		catch (ApiException e) {
+			log.error("failed sending notification message due to: " + e);
```
pkliczewski, line 37: Do you think it makes sense to propagate this to end user?

pkliczewski (commented):
overall looks good

hack/manifests/testing/parodos-patch.yaml
```diff
@@ -19,7 +19,7 @@ data:
   DATASOURCE_USERNAME: "parodos"
   DATASOURCE_PASSWORD: "parodos"
   SPRING_PROFILES_ACTIVE: "dev"
-  NOTIFICATION_SERVER_ADDRESS: "notification-service.default.svc.cluster.local"
+  NOTIFICATION_SERVER_URL: "http://notification-service.default.svc.cluster.local:8080"
```
masayag, line 22: @gciavarrini pls be aware of this change. 
this is changed by your open PR - https://github.com/parodos-dev/parodos/pull/376/files#diff-be2e48c86fd33f58e083bafa29850d09d6b6b5ac31c5e6371bd8d686a1bfdd3c

workflow-service/src/main/java/com/redhat/parodos/NotificationSender.java
masayag: IIUC, to propagate the exception a user can use `trySend` which throws.

workflow-service/src/main/java/com/redhat/parodos/NotificationSender.java
```diff
@@ -0,0 +1,46 @@
...
+	public NotificationSender(@Value("${notification.url}") String url,
+			@Value("${notification.auth.basic.user}") String user,
+			@Value("${notification.auth.basic.password}") String password) {
+		ApiClient apiClient = new com.redhat.parodos.notification.sdk.api.ApiClient().setBasePath(url).addDefaultHeader(
+				HttpHeaders.AUTHORIZATION,
+				"Basic " + Base64.getEncoder().encodeToString((user + ":" + password).getBytes()));
+		client = new NotificationMessageApi(apiClient);
+	}
+
+	public void send(NotificationMessageCreateRequestDTO message) {
```
masayag, line 32: pls add `@Override`

PR #392, opened 2023-05-31
Add task for deploying application manifests

masayag:
The PR adds a workflow task to read manifests from a given path and apply them to OCP cluster.
Once the resources are applied, the work context should be updated with the routers to the deployed application.

Fixes [FLPATH-411](https://issues.redhat.com/browse/FLPATH-411)

**Change type**
- [x] New feature
- [ ] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [x] Workflow Service
- [ ] Notification Service

**Checklist**

### DIFF
--- a/prebuilt-tasks/pom.xml
+++ b/prebuilt-tasks/pom.xml
@@ -19,6 +19,7 @@
         <azure-core-version>1.37.0</azure-core-version>
         <azure-identity-version>1.8.1</azure-identity-version>
         <azure-resourcemanager-version>2.24.0</azure-resourcemanager-version>
+        <fabric8.version>6.6.2</fabric8.version>
     </properties>
 
     <repositories>
@@ -158,6 +159,31 @@
             <version>${azure-resourcemanager-version}</version>
         </dependency>
         <!-- End libraries for Azure access -->
+
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>kubernetes-client</artifactId>
+            <version>${fabric8.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>openshift-client</artifactId>
+            <version>${fabric8.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>openshift-server-mock</artifactId>
+            <version>${fabric8.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- This specific version is needed for openshift server mock -->
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <version>4.11.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/deploy/DeployApplicationTask.java
+++ b/prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/deploy/DeployApplicationTask.java
@@ -1,0 +1,246 @@
+package com.redhat.parodos.tasks.deploy;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.redhat.parodos.workflow.context.WorkContextDelegate;
+import com.redhat.parodos.workflow.exception.MissingParameterException;
+import com.redhat.parodos.workflow.parameter.WorkParameter;
+import com.redhat.parodos.workflow.parameter.WorkParameterType;
+import com.redhat.parodos.workflow.task.BaseWorkFlowTask;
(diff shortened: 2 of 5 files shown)


prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
```diff
@@ -0,0 +1,132 @@
...
+		if (manifests.isEmpty()) {
+			return new DefaultWorkReport(WorkStatus.FAILED, workContext,
+					new RuntimeException("No manifest files found in path %s".formatted(manifestsPath)));
+		}
+
+		// Load the kubeconfig file and prepare the client
+		try (OpenShiftClient osClient = createOpenShiftClient(kubeconfig)) {
+			manifests.forEach(manifest -> applyManifest(osClient, namespace, manifest));
+		}
+		catch (Exception e) {
```
eloycoto, line 107: applyManifest throws a RuntimeException, no?

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
```diff
@@ -0,0 +1,132 @@
...
+	 * @return a {@link WorkReport} with the result of the execution
+	 */
+	@Override
+	public WorkReport execute(WorkContext workContext) {
+		final String kubeconfig;
+		final String manifestsPath;
+		final String namespace;
+
+		// Verify parameters are present
+		try {
```
eloycoto, line 74: KUBECONFIG is from the env variables inside the openshift, no?

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
```diff
@@ -0,0 +1,132 @@
...
+			kubeconfig = WorkContextDelegate.getRequiredValueFromRequestParams(workContext, KUBECONFIG);
+			manifestsPath = WorkContextDelegate.getRequiredValueFromRequestParams(workContext, MANIFESTS_PATH);
+			namespace = WorkContextDelegate.getRequiredValueFromRequestParams(workContext, NAMESPACE);
+		}
+		catch (MissingParameterException e) {
```
eloycoto, line 79: There is an ilelligalParameter exception if I recall correctly

eloycoto (changes requested):
Missing test.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
masayag: I don't think we should assume that the same cluster will be used for both running the workflow and also to deploy the application on.
I'd separate the two things. You can want to use cluster A as the target cluster to deploy the application and cluster B as the cluster on which you run the workflow service.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
masayag: https://github.com/parodos-dev/parodos/blob/d5eab829844b152d6f3d0e398fb23c10857a430c/parodos-model-api/src/main/java/com/redhat/parodos/workflow/context/WorkContextDelegate.java#L181

```
	public static String getRequiredValueFromRequestParams(WorkContext workContext, String key)
			throws MissingParameterException {
```

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
masayag: yes, I'll add a specific case for it after I'll find out what will be thrown in the case of  a failure to create a resource or for reading a resource after it was created.
there is a metadata object returned by this call that needs to be analyzed and to store the router information on the context. this PR isn't complete yet.

masayag:
> Missing test.

@eloycoto before writing tests, I wanted to make sure the expectation for the input parameters is correct

eloycoto:
> @eloycoto before writing tests, I wanted to make sure the expectation for the input parameters is correct

Overall LGTM, maybe I'll move the set<Path> manifest to a public function; I can think of extending the class without parameters and deploying the manifest directly into it.

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
eloycoto: :+1:

prebuilt-tasks/src/main/java/com/redhat/parodos/tasks/application/DeployApplicationTask.java
masayag: added tests to cover the various cases

masayag:
> > @eloycoto before writing tests, I wanted to make sure the expectation for the input parameters is correct
> 
> Overall LGTM, maybe I'll move the set manifest to a public function; I can think of extending the class without parameters and deploying the manifest directly into it.

feel free to modify the code as you see once it is merged to satisfy the needs of the next evolution of it, or let me know what you'd like me to change.

pkliczewski:
@ydayagi it would be good to use this client instead

PR #399, opened 2023-06-06
Improve waiting strategy for Integration tests

masayag:
Instead of relying on okhttp async implementation, we use executor service to invoke recurring requests to the service till completion.
This is designed to decrease the CPU load even lower while maintaining more readable and easier-to-maintain code.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story (FLPATH-xxxx)*:
Fixes #[FLPATH-399](https://issues.redhat.com/browse/FLPATH-399)

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Unit tests
- [x] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [x] Workflow Service
- [ ] Notification Service

**Checklist**

### DIFF
--- a/sdk-utils/src/main/java/com/redhat/parodos/sdkutils/RetryExecutorService.java
+++ b/sdk-utils/src/main/java/com/redhat/parodos/sdkutils/RetryExecutorService.java
@@ -1,0 +1,87 @@
+package com.redhat.parodos.sdkutils;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+public class RetryExecutorService<T> implements AutoCloseable {
+
+	private final ExecutorService executor;
+
+	public RetryExecutorService() {
+		executor = Executors.newFixedThreadPool(1);
+	}
+
+	/**
+	 * Submit a task to the executor service, retrying on failure until the task succeeds
+	 * @param task The task to submit
+	 * @return The result of the task
+	 */
+	public T submitWithRetry(Callable<T> task) {
+		// @formatter:off
+		return submitWithRetry(task, () -> {}, () -> {}, 10 * 60 * 1000, 5000);
+		// @formatter:on
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


sdk-utils/src/main/java/com/redhat/parodos/sdkutils/SdkUtils.java
```diff
@@ -57,7 +51,7 @@ public static ApiClient getParodosAPiClient()
 		}
 
 		int port = Integer.parseInt(serverPort);
-		if (port <= 0 && port > 65535) {
+		if (port <= 0 || port > 65535) {
```
masayag, line 54: this isn't related to this PR, but fixes a bug since this if used to return always false.

sdk-utils/src/main/java/com/redhat/parodos/sdkutils/RetryExecutorService.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * Submit a task to the executor service, retrying on failure until the task succeeds
+	 * @param task The task to submit
+	 * @return The result of the task
+	 */
+	public T submitWithRetry(Callable<T> task) {
+		// @formatter:off
+		return submitWithRetry(task, () -> {}, () -> {}, 10 * 60 * 1000, 5000);
```
masayag, line 25: @gciavarrini what should be the timeout? was it used to be 1 minute in previous implementation?

gciavarrini: In the bash implementation there were 100 retries.
I chose 1 minute because it seemed like a reasonable time interval and I didn't receive any feedback on this.
TBH, I'm not sure which is the best timeout value :)

sdk-utils/src/main/java/com/redhat/parodos/sdkutils/SdkUtils.java
gciavarrini: you're right. Fixes even also https://github.com/parodos-dev/parodos/pull/376/commits/e837ae60a4b119f394791728aa4f3c96dffcf3cf

PR #402, opened 2023-06-06
FLPATH-405: etag & if-not-modified

RichardW98:
return 304 if the value of  if-not-modified in request header is equal to etag
Fixes #405

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [ ] Workflow Service
- [ ] Notification Service

**Checklist**

### DIFF
--- a/workflow-service/src/test/java/com/redhat/parodos/project/controller/ProjectControllerTest.java
+++ b/workflow-service/src/test/java/com/redhat/parodos/project/controller/ProjectControllerTest.java
@@ -108,6 +108,20 @@ public class ProjectControllerTest extends ControllerMockClient {
 	}
 
 	@Test
+	public void testListProjects_when_etagEquals_then_return_notModified() throws Exception {
+		ProjectResponseDTO project1DTO = createSampleProject(PROJECT_NAME_1);
+		ProjectResponseDTO project2DTO = createSampleProject(PROJECT_NAME_2);
+		when(projectService.getProjects()).thenReturn(List.of(project1DTO, project2DTO));
+		String etagValue = String.valueOf(List.of(project1DTO, project2DTO).hashCode());
+		// When
+		mockMvc.perform(this.getRequestWithValidCredentials("/api/v1/projects/").header("If-None-Match",
+				"\"" + etagValue + "\"")).andExpect(MockMvcResultMatchers.status().isNotModified());
+
+		// Then
+		verify(projectService, times(1)).getProjects();
+	}
+
+	@Test
 	public void testGetProjectsWithInvalidCredentials() throws Exception {
 		// When
 		mockMvc.perform(this.getRequestWithInValidCredentials("/api/v1/projects/"))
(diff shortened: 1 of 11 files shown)


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkFlowStatusResponseDTO.java
```diff
@@ -45,6 +46,6 @@ public class WorkFlowStatusResponseDTO {
 	private WorkStatus status;
 
 	@JsonInclude(JsonInclude.Include.NON_EMPTY)
-	private List<WorkStatusResponseDTO> works;
+	private transient List<WorkStatusResponseDTO> works;
```
masayag, line 49: if this one is `transient`, shouldn't the annotation be changed to `@JsonIgnore` or should it remain serialized to json?

RichardW98: we need it to be serialized.  sonarlint requires jsonignore or transient, so I have to pick one

RichardW98: this is dto, so I think transient won't afffect

workflow-service/src/test/java/com/redhat/parodos/project/controller/ProjectControllerTest.java
```diff
@@ -206,7 +220,7 @@ public void testRemoveUsersFromProject_WithEmptyUsersList() throws Exception {
 
 	ProjectResponseDTO createSampleProject(String name) {
 		ProjectResponseDTO responseDTO = new ProjectResponseDTO();
-		responseDTO.setId(UUID.randomUUID());
+		responseDTO.setId(UUID.fromString("af7c1fe6-d669-414e-b066-e9733f0de7a8"));
```
gciavarrini, line 223: Some tests use this method.
WDYT to set the "hard coded" `Id` in  `testListProjects_when_etagEquals_then_return_notModified` test instead?

RichardW98: I was expecting the hash of the same UUID string should always return the same value. but different java version may have different result, I will rollback to randomUUID

gciavarrini: If we want to track this, maybe we can add another unit tests. So we can be sure that each generated id will not change

masayag (approved):

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkFlowStatusResponseDTO.java
RichardW98: i rolled back all the serialzable implement.   obj.hashcode() doesn't need it

PR #410, opened 2023-06-07
Release to GH and Publish Maven Packages using github action

rgolangh:
This action will push the packages when a tag v* is pushed.

To initiated it:
    ```
    make release-tag-push
    ```

That would promote the version of the pom ,tag with v*, and push the tag
to origin. A github action would kick in and:

- create a github release
- push modules to maven
- build container images
- push container images
- bump the next version
- push the a commit with next

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,58 @@
+name: Full Release on Tag Push
+on:
+  push:
+    tags:
+      - v**
+jobs:
+  GitHub-Release:
+    runs-on: ubuntu-latest
+    steps:
+       steps:
+        - run: gh release create ${{  github.ref_name }} --generate-notes
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  Maven-Publish:
+    runs-on: ubuntu-latest
+      - uses: actions/checkout@v3
+
+      - name: "Set up Maven Central Repository"
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          server-id: ossrh
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+
+      - name: "Publish package to maven repo"
+        run: mvn --batch-mode deploy
+        env:
+          MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+          MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
+
+  Container-Publish:
+    - name: "Build container images"
+      run: |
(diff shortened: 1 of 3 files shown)


rgolangh:
@pkliczewski  ptal
Someone with access to the secrets  in the org or project level should set the 2 secrets for `OSSRH_USERNAME` and `OSSRH_TOKEN`

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Set up Maven Central Repository
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          server-id: ossrh
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+      - name: Publish package
+        run: make release-all
```
masayag, line 19: I think this one is tricky: we shouldn't invoke `release-all` here.
`release-all` creates the git-commit of the tag which should trigger this workflow after it is merged.
sounds like an endless cycle.

If this one should only publish maven resources - then we invoke only the `make release` or `mvn deploy -P release` and remove that option from `make release-all`

@pkliczewski WDYT?

rgolangh: this flow is triggered only on a github release , not on a tag push

masayag: oh, so my bad here.
you are correct and that was explained in the commit message.

masayag:
After this workflow is merged, perhaps we can add a badge to the project main page with the release status?

.github/workflows/publish.yaml
rgolangh: I hope this all would work :pray:

eloycoto:
I would prefer to keep this PR on hold until Monday, I need to do the release and I'm not sure about all the implications so I like to wait and don't mess up the next release.

Makefile
```diff
@@ -192,8 +192,13 @@ git-tag: ## tag commit and prepare for image release
 	git tag -a $(TAG) -m "$(TAG)"
 	$(eval TAG=$(TAG))
 
+# Officially release using local machine. Deprecated.
 release-all: update-version release git-release git-tag build-images tag-images push-images bump-version install bump-git-commit
 
+# Officially release using github automation
+release-tag-push: update-version git-release git-tag
+	git push origin v$(RELEASE_VERSION)
```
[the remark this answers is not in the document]
pkliczewski: PR for a release is created once we run `mvn clean install -P release` which already pushes everything to maven central. After the release we need to update with new SNAPSHOT release.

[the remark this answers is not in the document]
rgolangh: created automatically?

PR #411, opened 2023-06-08
[FLPATH-424] Return error message with status

gabriel-farache:
This PR adds the error message contained in the `Throwable` of the `WorkReport` into the `WorkflowExecution` and saves it to the DB in order to return it when the `/status` endpoint is requested; the message is then sent along with the status

**Which issue(s) this PR fixes** 
`fixes #FLPATH-424` 

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [x] Workflow Service
- [ ] Notification Service

**Checklist**
- [x] Relevant issues have been referenced.

### DIFF
--- a/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkStatusResponseDTO.java
+++ b/workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkStatusResponseDTO.java
@@ -47,6 +47,8 @@ public class WorkStatusResponseDTO {
 
 	private WorkStatus status;
 
+	private String message;
+
 	@JsonProperty("works")
 	@JsonInclude(JsonInclude.Include.NON_EMPTY)
 	private List<WorkStatusResponseDTO> works;
--- a/workflow-service-sdk/api/openapi.yaml
+++ b/workflow-service-sdk/api/openapi.yaml
@@ -1332,17 +1332,22 @@ components:
           - null
           - null
           name: name
+          message: message
           type: TASK
           status: FAILED
         - works:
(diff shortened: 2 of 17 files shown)


workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkStatusResponseDTO.java
```diff
@@ -47,6 +47,8 @@ public class WorkStatusResponseDTO {
 
 	private WorkStatus status;
 
+	private String message;
```
masayag, line 50: `@JsonInclude(JsonInclude.Include.NON_EMPTY)`  ?

masayag:
overall, LGTM, but have one question about whether should we include a message in response if it is empty

workflow-service/src/main/java/com/redhat/parodos/workflow/execution/dto/WorkStatusResponseDTO.java
gabriel-farache: The API contract/spec defines that the response contains a `message` field, this field shall be present for parodos to fulfil the contract.
This way, the API consumer will not have to worry about non-existing key when processing the response

RichardW98: I think it can be optional. we can address it later

PR #433, opened 2023-06-15
Pass context params from an execution to a workflow execution

masayag:
Add invokingExecutionId to WorkflowRequestDTO to support context passing
    
Enable using arguments from another execution to use its execution level parameters.
    
This could be used when an assessment workflow gathers information and stores it in the context and then when invoking a follow-up workflow from the options of that assessment then the requester can pass the execution ID of the assessment.
    
The data flow looks like that:
```    
    Start -> Assessment flow -> addParameter "foo": "bar" to WorkContext -> End with option 1
    
    Start -> Workflow from option 1 -> use Workcontext param "foo" -> work -> End
```    

The invoking execution ID isn't stored in the DB, it is just a request argument passed in the POST request for invoking a workflow.
    
Signed-off-by: Moti Asayag <masayag@redhat.com>

[FLPATH-458](https://issues.redhat.com//browse/FLPATH-458)


pkliczewski (commented):
/lgtm

RichardW98 (approved):

eloycoto (approved):

PR #445, opened 2023-06-16
feat: add release manual action

eloycoto:
This is not to replace the current work, is to avoid the pain suffered over the last releases where things didn't work as good as expected.

Manual workflow, triggered manually, to release it. As I said, not perfect, but want to reduce the current friction.


masayag (commented):
/lgtm

pkliczewski (approved):

PR #447, opened 2023-06-19
feat: append manifests to release

eloycoto:
When a new release is created, this GitHub action will automatically push to the release the manifests with the correct images for Parodos and backstage with the plugin.


pkliczewski (commented):
/lgtm

PR #460, opened 2023-06-28
FLPATH-480:recommend option & return workflow type in workflowResponseDto

RichardW98:
recommend option & return workflow type in workflowResponseDto
Fixes #480

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [ ] Workflow Service
- [ ] Notification Service

**Checklist**

### DIFF
--- a/workflow-examples/src/main/java/com/redhat/parodos/examples/prebuilt/migrationtoolkit/MigrationAssessmentWorkflow.java
+++ b/workflow-examples/src/main/java/com/redhat/parodos/examples/prebuilt/migrationtoolkit/MigrationAssessmentWorkflow.java
@@ -68,13 +68,6 @@ public class MigrationAssessmentWorkflow {
 	}
 
 	@Bean
-	WorkFlowOption move2kube() {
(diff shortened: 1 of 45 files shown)


workflow-examples/src/main/java/com/redhat/parodos/examples/prebuilt/migrationtoolkit/MigrationAssessmentWorkflow.java
```diff
@@ -67,13 +67,6 @@ ProcessAnalysisTask processAnalysisTask(WorkFlowOption move2kube, WorkFlowOption
 				notifier);
 	}
 
-	@Bean
```
masayag, line 70: not sure if this is the cause of the integration tests failure which complains about unsatisfied dependency - pls have a look at the build log failure.

RichardW98: yes the bean dependency 'move2kube' is not loaded in advance

PR #467, opened 2023-07-04
feat: backstage manifests using single namespace

eloycoto:
Instead of using backstage in other namespace, just don't use, this will allow  to deploy multiple instances in different namespaces:

Like:
```
kubectl kustomize hack/manifests/backstage/ | kubectl apply -n differentOne -f -
```

Or in default:

```
kubectl kustomize hack/manifests/backstage/ | kubectl apply -n default -f -
```

Fix FLPATH-511


hack/manifests/base/deployment.yaml
```diff
@@ -60,5 +60,5 @@ spec:
           httpGet:
             path: /actuator/health
             port: 8080
-          initialDelaySeconds: 25
+          initialDelaySeconds: 40
```
masayag, line 63: is this related to the PR? meaning, does not having a specified namespace require more time or this is needed for other issue?

masayag (approved):

PR #470, opened 2023-07-06
separate workflow-examples from workflow-service

ydayagi:
build workflow-service image without workflow-examples
build an image with workflow-examples for integration tests

[FLPATH-510](https://issues.redhat.com//browse/FLPATH-510) https://issues.redhat.com/browse/FLPATH-510

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -111,6 +111,11 @@ jobs:
       - name: Upload image
         uses: ishworkh/docker-image-artifact-upload@v1
         with:
+          image: "docker-compose_examples-service:latest"
+
+      - name: Upload image
+        uses: ishworkh/docker-image-artifact-upload@v1
+        with:
           image: "docker-compose_workflow-service:latest"
 
       - name: Upload image
@@ -175,6 +180,11 @@ jobs:
         with:
           image: "docker-compose_workflow-service:latest"
 
+      - name: Download images
+        uses: ishworkh/docker-image-artifact-download@v1
+        with:
+          image: "docker-compose_examples-service:latest"
+
       - name: Load images inside kind
         run: |
           make push-images-to-kind
@@ -185,6 +195,7 @@ jobs:
           kubectl wait --timeout=600s --for=condition=Ready pods --all -n default || {
             echo "Timeout waiting for Parodos services to be ready";
             kubectl get pods --all-namespaces;
+            kubectl describe pods
             echo "************ workflow-service logs ***************";
             kubectl logs -l app=workflow-service --tail=40;
             echo "************ notification-service logs ***************";
--- a/hack/manifests/testing/kustomization.yaml
+++ b/hack/manifests/testing/kustomization.yaml
@@ -14,6 +14,7 @@ patches:
 
 images:
 - name: quay.io/parodos-dev/workflow-service:main
+  newName: quay.io/parodos-dev/examples-service
   newTag: test
 - name: quay.io/parodos-dev/notification-service:main
   newTag: test
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ MAVEN ?= /usr/bin/mvn
 ORG=quay.io/parodos-dev/
 WORKFLOW_SERVICE_IMAGE=workflow-service
 NOTIFICATION_SERVICE_IMAGE=notification-service
+EXAMPLES_SERVICE_IMAGE=examples-service
 
 # get version from pom
 VERSION = $(shell sed -n "s/<revision>\(.*\)<\/revision>/\1/p" $(PWD)/pom.xml | sed 's/\ *//g')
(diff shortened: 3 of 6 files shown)


hack/manifests/base/deployment.yaml
```diff
@@ -43,12 +43,19 @@ spec:
       labels:
         app: workflow-service
     spec:
+      volumes:
+        - name: user-jar
+          persistentVolumeClaim:
+            claimName: user-jar
```
masayag, line 49: let's rename it to `workflows-jar` or `user-workflows-jar`

hack/manifests/testing/copy-to-pvc.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: user-jar
```
masayag, line 4: here as well: `workflows-jar`

.github/workflows/test.yml
```diff
@@ -175,16 +180,31 @@ jobs:
...
+        with:
+          image: "docker-compose_workflow-examples:latest"
+
       - name: Load images inside kind
         run: |
           make push-images-to-kind
 
+      - name: Copy workflow-examples to PVC
+        run: |
+          kubectl create -f hack/manifests/testing/copy-to-pvc.yaml
```
masayag, line 194: @eloycoto at this point, do we have the workflow jar available in the downloaded artifacts?

if we have it, we can replace packaging the workflow-jar as image and instead of using pod to copy form one pod to another PVC with `kubectl cp` option
https://www.middlewareinventory.com/blog/kubectl-cp-example/

.github/workflows/test.yml
```diff
@@ -175,16 +180,31 @@ jobs:
...
+            exit 1;
+          }
+
       - name: Deploy
         run: |
           kubectl kustomize hack/manifests/testing| kubectl apply -f -
           kubectl wait --timeout=600s --for=condition=Ready pods --all -n default || {
             echo "Timeout waiting for Parodos services to be ready";
             kubectl get pods --all-namespaces;
+            kubectl describe pods
```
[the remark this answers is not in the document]
ydayagi: When I have a failed pod I use 'decribe' more than logs. Logs are for application errors. These are easy to debug. Other errors such as security, volumes etc are only covered by 'describe'. If u don't like i will change

hack/manifests/testing/copy-to-pvc.yaml
ydayagi: will fix

.github/workflows/test.yml
ydayagi: @masayag I tested it. it works. I don't think that an image is a problem here. you will still need the extra pod. Because 'kubectl cp' only copies to a container. but with an image it is less intrusive to the way pods work. compared to 'kubectl cp'. and you get all the logic in the pod yaml instead of a pod and a 'kubectl cp' command.

ydayagi: notice that the image is only used locally. it is not pushed to Quay.

hack/manifests/base/deployment.yaml
ydayagi: fixed

hack/manifests/testing/copy-to-pvc.yaml
ydayagi: fixed

eloycoto (commented):
The code looks good, but I do not think that it's the best way to present this to the users. At the end, this is not composable at all. 

I would like to be the jars part of the Dockerfile, so the user can push the Jar inside the Dockerfile and keep track of IDS and versions. 

I cannot see how a rollback in a version will work here or how can a user can register 10 different jars to compose user flows. 

Not sure if this is the right direction.

ydayagi:
> The code looks good, but I do not think that it's the best way to present this to the users. At the end, this is not composable at all.
> 
> I would like to be the jars part of the Dockerfile, so the user can push the Jar inside the Dockerfile and keep track of IDS and versions.
> 
> I cannot see how a rollback in a version will work here or how can a user can register 10 different jars to compose user flows.
> 
> Not sure if this is the right direction.

@eloycoto @masayag @pkliczewski 
Eloy please suggest another solution.
Regarding, rollback. I don't see how putting the JARs in the Dockerfile helps. As long as u have two separate entities/deliverables: user JARs and our JARs, you will have a problem.
Supporting multiple JARs is possible. it is just a decision.

hack/manifests/testing/kustomization.yaml
```diff
@@ -14,6 +14,7 @@ patches:
 
 images:
 - name: quay.io/parodos-dev/workflow-service:main
+  newName: quay.io/parodos-dev/integration-service
```
masayag, line 17: Is there a need to create this repository on quay?

ydayagi: it is only a tag. it is never pushed to quay

eloycoto: should main added here, no?

eloycoto (changes requested):
I'm not sure if the push-image thing is implemented, I do not think so. 

At the other hand, maybe the update on hack/manifest is needed, to use the image in all manifest, if not only the workflow-service without workflows examples in there. 

I would check all .github/actions/ and hack/manifest/

hack/manifests/testing/kustomization.yaml
ydayagi: @eloycoto what do u mean?

eloycoto: Should be like this:

```
- name: quay.io/parodos-dev/workflow-service:main
  newTag: test
- newName: quay.io/parodos-dev/integration-service:main
  newTag: test
- name: quay.io/parodos-dev/notification-service:main
  newTag: test
```

ydayagi: no. this is how it should be. and it works

ydayagi:
> I'm not sure if the push-image thing is implemented, I do not think so.
> 
> At the other hand, maybe the update on hack/manifest is needed, to use the image in all manifest, if not only the workflow-service without workflows examples in there.
> 
> I would check all .github/actions/ and hack/manifest/

@masayag @eloycoto @pkliczewski I don't understand what is needed. the required changes for building an image for integration tests were made. it does not effect release.

pkliczewski:
@eloycoto what do you think?

eloycoto (commented):
LGTM, but:

@gabriel-farache, maybe we should update the image in the operator default one? 
@ydayagi  did you create the image in the quay.io?
 
I'll notify users to use examples-service instead of workflow-service. cc @RichardW98

gabriel-farache:
@eloycoto 
> @gabriel-farache, maybe we should update the image in the operator default one?

When we release the new version of workflow-service, the `stable` image will not have the examples, as there are in a separate image now; so do you mean we should add the example image to the operator? 
BTW how that split will reflect in the manifest we release?

pkliczewski:
@gabriel-farache @eloycoto what do you suggest to do to finalize this PR?

eloycoto:
Quay container created:
https://quay.io/repository/parodos-dev/examples-service

PR #473, opened 2023-07-10
Fix method name to `getConcreteImplementation`

gciavarrini:
Fix method name to `getConcreteImplementation`

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Unit tests
- [ ] Integration tests
- [ ] CI
- [ ] Documentation
- [ ] Auto-generated SDK code

**Impacted services**
- [ ] Workflow Service
- [ ] Notification Service

**Checklist**


workflow-examples/src/test/java/com/redhat/parodos/examples/base/BaseAssessmentTaskTest.java
```diff
@@ -19,13 +19,13 @@ public abstract class BaseAssessmentTaskTest {
...
 	public void setUp() {
-		this.baseAssessmentTask = getConcretePersonImplementation();
+		this.baseAssessmentTask = getConcreteImplementation();
 	}
 
 	/**
 	 * @return new instance of non-abstract class extending BaseInfrastructureWorkFlowTask
 	 */
-	protected abstract BaseAssessmentTask getConcretePersonImplementation();
+	protected abstract BaseAssessmentTask getConcreteImplementation();
```
masayag, line 28: if this is designed only for testing, shouldn't the name be related to it?
perhaps `getTaskUnderTest` or something like this?

something is a bit off when reading in the same definition both the word `abstract`, `concrete`, and `implementation` 

the perhaps instead of describe how this is done, the method should describe what it should do.

gciavarrini: Can you please check if the description is ok now?

masayag: The description is good.
But something is strange: I see both getTaskUnderTest and getConcreteImplementation in the PR.

gciavarrini: You're right. I think it's ok now

masayag (approved):

PR #477, opened 2023-07-17
Implement integration tests for prebuilt tasks

ydayagi:
3 workflow tasks are now tested: kubeapi, rest, jdbc.
their code was originally in workflow-examples and now moved to integration-tests
REST simply makes a GET request to the workflow-service
JDBC performs all its operations in the Postgres DB.
kubeapi uses the kind cluster kubeconfig. we change the URL so it is reachable from within the workflow-service pod. we also create a test namespace for the kubeapi test resouces.

[flpath-401](https://issues.redhat.com//browse/flpath-401) https://issues.redhat.com/browse/FLPATH-401

 Signed-off-by: Yaron Dayagi <ydayagi@redhat.com>


integration-tests/src/test/java/com/redhat/parodos/integration/prebuilt/RestTest.java
```diff
@@ -0,0 +1,59 @@
...
+		assertNotNull(workFlowResponseDTO.getWorkStatus());
+		assertEquals(WorkFlowExecutionResponseDTO.WorkStatusEnum.IN_PROGRESS, workFlowResponseDTO.getWorkStatus());
+
+		WorkFlowStatusResponseDTO workFlowStatusResponseDTO = WorkFlowServiceUtils.waitWorkflowStatusAsync(workflowApi,
+				workFlowResponseDTO.getWorkFlowExecutionId());
+
+		assertNotNull(workFlowStatusResponseDTO);
+		assertNotNull(workFlowStatusResponseDTO.getWorkFlowExecutionId());
+		assertNotNull(workFlowStatusResponseDTO.getStatus());
+		assertEquals(WorkFlowStatusResponseDTO.StatusEnum.COMPLETED, workFlowStatusResponseDTO.getStatus());
```
gciavarrini, line 56: According to the project rules, we should use `assertThat().isNotNull`, `assertThat().isEqualTo`, etc.

workflow-examples/src/main/java/com/redhat/parodos/examples/JdbcWorkFlowConfiguration.java
gciavarrini, line 1: May I ask why this and the other config in this folder are still necessary?

integration-tests/src/test/java/com/redhat/parodos/integration/prebuilt/RestTest.java
ydayagi: this piece of code was copied from existing integration tests. so these rules probably do not apply there

workflow-examples/src/main/java/com/redhat/parodos/examples/JdbcWorkFlowConfiguration.java
ydayagi: this is the workflow definition. like all our workflows. only the tests/execution/client is in integration tests.

RichardW98 (approved):

PR #482, opened 2023-07-26
feat: release tag latest version with stable tag

eloycoto:
Backstage Parodos is still under different version tracking, so not been added here yet!


pkliczewski (commented):
/lgtm

